PR #82, opened 2023-09-05
[svg] / [design-system] - Update explicit types, and added `eslint` config to `svg`

nebula-aac:
**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/packages/components/src/accordion.tsx
+++ b/packages/components/src/accordion.tsx
@@ -1,6 +1,6 @@
-import { AccordionProps, Accordion as MuiAccordion } from '@mui/material';
+import { Accordion as MuiAccordion, type AccordionProps } from '@mui/material';
 import React from 'react';
 
-export function Accordion(props: AccordionProps) {
+export function Accordion(props: AccordionProps): JSX.Element {
   return <MuiAccordion {...props} />;
 }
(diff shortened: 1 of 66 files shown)


aabidsofi19:
why did we change it to JSX.Element  (curious)

nebula-aac:
The updates comes from `eslint` rule

Here is an example of error when you run `eslint`.

```bash
/workspaces/sistent/packages/components/src/base/Accordion/accordionsummary.tsx
  4:8  error  Missing return type on function  @typescript-eslint/explicit-function-return-type
```

To satisfy the rule, we can explicitly label `JSX.Element` which means each of the component is returning an `JSX.Element`

We can either turn this rule off to make it less explicit, and allow some flexibility in using the components, or leave this rule on, and make sure that each of the component is returning an `JSX.Element`. This will allow Typescript to perform better type checking and also allow everyone else who is contributing to know this this is a particular type.

I'm open to either way.

nebula-aac:
@aabidsofi19 Can you review again and see if this can be merged in.

I have more that I want to add, and don't want to create multiple PRs and rebasing the PRs if I can avoid it.

aabidsofi19:
@nebula-aac sorry ,  i forgot to review some of the prs  ..

aabidsofi19:
@nebula-aac  i think the return type should be React.FC rather than JSX.Element , because it is a react component as it takes props

leecalcote:
Merge conflicts cropped up...

nebula-aac:
I'm creating a different PR, closing this one.

Issue #73, opened 2023-08-31
[components] Create custom search bar


nebula-aac:
In Meshery UI, we have created a search bar, but it is a mixture of Material UI v4 and MUI v5.

Here is the file for reference: https://github.com/meshery/meshery/blob/master/ui/components/searchcommon.js

There is a need to create a custom search bar in the `components` package to be reused.

At this time, there is not a default `ThemeProvider` in this repo, so at this, it would be great to create a custom search bar from the implementation in Meshery UI and display this in the Storybook for now.

---

**[Contributor Guide](https://docs.meshery.io/project/contributing)**

- 🎨 Wireframes and [designs for Layer5 site](https://www.figma.com/file/5ZwEkSJwUPitURD59YHMEN/Layer5-Designs) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)

harshal2030:
Hey, can I work on this

nebula-aac:
This is done

PR #85, opened 2023-09-05
[components] created `searchbar` and `tooltip`

nebula-aac:
**Notes for Reviewers**

This PR fixes https://github.com/layer5io/sistent/issues/72 and https://github.com/layer5io/sistent/issues/73

In Meshery UI and Meshery Cloud and any others that uses MUI, we have a reusable tooltip but we're duplicating work in this.

For the `search bar`, we have this done in Meshery UI but it is not a complete replica of the one from Meshery UI, but it does takes most of the basic properties of what we're using Meshery UI.

These two components are also being used in the Design System.


aabidsofi19:
@nebula-aac  we should rename the `meshery` folder . maybe `base` for components are to be customized and `core` for the components that are more or less completely  styled , also the component name (filename) should not contain the name of folder then.

nebula-aac:
This has been updated to rename the directories and the names of the custom components @aabidsofi19

nebula-aac:
@aabidsofi19 can you review here again?

aabidsofi19 (approved):
Looks great

PR #178, opened 2023-10-25
[components] Create custom modal

### LINKED ISSUE #74: [components] Create custom modal
Requirements:

- [ ] Create a custom `Modal` using the custom `Dialog` or the base `Dialog` components exported from the `components` library.

---

**[Contributor Guide](https://docs.meshery.io/project/contributing)**

- 🎨 Wireframes and [designs for Layer5 site](https://www.figma.com/file/5ZwEkSJwUPitURD59YHMEN/Layer5-Designs) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)

coder12git:
**Notes for Reviewers**

This PR fixes #74

### DIFF
--- a/packages/components/src/custom/Modal/modal.tsx
+++ b/packages/components/src/custom/Modal/modal.tsx
@@ -1,0 +1,26 @@
+import { FC, ReactNode } from 'react';
+import {
+  StyledDialog,
+  StyledDialogActions,
+  StyledDialogContent,
+  StyledDialogTitle
+} from '../Dialog';
+
+interface MesheryModalProps {
+  open: boolean;
+  content: ReactNode;
+  title: string;
+  actions?: ReactNode;
+}
+
+const MesheryModal: FC<MesheryModalProps> = ({ open, content, title, actions }) => {
+  return (
+    <StyledDialog fullWidth maxWidth="md" open={open}>
+      <StyledDialogTitle>{title}</StyledDialogTitle>
+      <StyledDialogContent>{content}</StyledDialogContent>
+      <StyledDialogActions>{actions}</StyledDialogActions>
+    </StyledDialog>
+  );
+};
+
+export { MesheryModal as StyledModal };
--- a/packages/components/src/custom/Modal/index.ts
+++ b/packages/components/src/custom/Modal/index.ts
@@ -1,0 +1,1 @@
+export { StyledModal } from './modal';


packages/components/src/custom/Modal/modal.tsx
```diff
@@ -0,0 +1,26 @@
...
+import {
+  StyledDialog,
+  StyledDialogActions,
+  StyledDialogContent,
+  StyledDialogTitle
+} from '../Dialog';
+
+interface MesheryModalProps {
+  open: boolean;
+  content: ReactNode;
```
aabidsofi19, line 11: if content is going to be a react node , we should accept it as children .

aabidsofi19 (changes requested):

packages/components/src/custom/Modal/modal.tsx
coder12git: i agree, but i used react node here, to specify that any type of content can be render here, in case if we pass children here then also we need to specify {children: ReactNode }, to specify that it acn hold any react element

aabidsofi19: ok if this is how it is already used , then great lets stay with that .

leecalcote:
@Yashsharma1911 is this in following with the reusable modals that we have in Cloud and Meshery UI currently?

packages/components/src/custom/Modal/modal.tsx
```diff
@@ -0,0 +1,26 @@
...
+  title: string;
+  actions?: ReactNode;
+}
+
+const MesheryModal: FC<MesheryModalProps> = ({ open, content, title, actions }) => {
+  return (
+    <StyledDialog fullWidth maxWidth="md" open={open}>
+      <StyledDialogTitle>{title}</StyledDialogTitle>
```
Yashsharma1911, line 19: @coder12git below image shows how our custom modal looks like currently, can you make sure the things I mentioned with red square in this image are added in this modal as these are going to be reusable

For reference, you can look at [Modal.js](https://github.com/meshery/meshery/blob/master/ui/components/Modal.js) file in meshery

Screenshot 2023-11-01 at 7 20 46 AM

It might be possible that gradient is not updated in meshery modal, so here is style for header background
```
  boxShadow: "inset 0px -1px 3px 0px rgba(0,0,0,0.2)",
  background: "linear-gradient(90deg, rgba(57,102,121,1) 0%, rgba(84,129,148,1) 100%)",
  filter:
    "progid:DXImageTransform.Microsoft.gradient(startColorstr='#396679',endColorstr='#548194',GradientType=1)",
```

for footer

```
  backgroundColor: "#396679",
  boxShadow: "inset 0px 3px 5px 0px rgba(0,0,0,0.25)"
```

`Left icon` and `helperIcon` are going to be conditionally rendered on the basis of props given to this modal, see this code for reference https://github.com/meshery/meshery/blob/2e31041178f3d666e8bad6f9f89bdd4ec21ab609/ui/components/Modal.js#L217

Yashsharma1911 (changes requested):

packages/components/src/custom/Modal/modal.tsx
coder12git: Sure, I will do the changes!

leecalcote:
@coder12git changes still pending?

Yashsharma1911:
> Is this PR still need to be reviewed or not?

Thanks for asking @nebula-aac , I think @aabidsofi19 also made a new reusable modal which we will use he might create a new PR for that, after that I wonder we can close this PR

PR #271, opened 2023-11-09
Mui Table some fixes to use column visibility

sudhanshutech:
**Notes for Reviewers**

We having a custom column visibility to check or uncheck columns to show/hide form view that is not having any effect , this change might fix that


sudhanshutech:
@nebula-aac  i need to make tableCols and updateCols as prop not state , but when doing giving eslint error

sudhanshutech:
@nebula-aac i have updated, just confirming there are two directories for responsive table, i think i had made changes to the right one

PR #403, opened 2023-12-05
Add `fill` as prop 

sudhanshutech:
**Notes for Reviewers**

This PR fixes #

Here we can pass custom fill color to icon we want accordingly where this component will be used


sudhanshutech:
@nebula-aac i am trying if this works then i would be able to pass theme reference to the prop while using in meshery

PR #465, opened 2024-01-17
[Components] Add sistent naming convention for our components

sudhanshutech:
**Notes for Reviewers**

This PR fixes #

For being consistent we can use the sistent naming convention instead of mui for exporting these  theme based components.

### DIFF
--- a/packages/components/src/theme/components/checkbox.modifier.ts
+++ b/packages/components/src/theme/components/checkbox.modifier.ts
@@ -1,7 +1,7 @@
 import { Components, Theme } from '@mui/material';
 import { CHARCOAL, connected, darkSlateGray } from '../colors';
 
-export const MuiCheckbox: Components<Theme>['MuiCheckbox'] = {
+export const SistentCheckbox: Components<Theme>['MuiCheckbox'] = {
   styleOverrides: {
     root: {
       color: 'transparent',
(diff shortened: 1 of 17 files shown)


sudhanshutech:
> So we're only just naming the custom components, not the exported components to be used in Meshery?

They might because i think these styled component will only play major role when wrapped with sistent provider right!.
The naming is for only to keep our sistent codebase looks like its our own components and colors.

packages/components/src/theme/components/checkbox.modifier.ts
```diff
@@ -1,7 +1,7 @@
 import { Components, Theme } from '@mui/material';
 import { CHARCOAL, connected, darkSlateGray } from '../colors';
 
-export const MuiCheckbox: Components<Theme>['MuiCheckbox'] = {
+export const SistentCheckbox: Components<Theme>['MuiCheckbox'] = {
```
aabidsofi19, line 4: @sudhanshutech  using sistent as prefix is redundant ( as they will be scoped under sistent ) , I would suggest use `Base-<Component>` for unstyled / customizable component and `<Functionality-<Component>` for customized and ready to use ones  , @nebula-aac  whats your opinion ?

aabidsofi19 (changes requested):

aabidsofi19:
@sudhanshutech  @nebula-aac  I would suggest using our custom convention for both private and public components

packages/components/src/theme/components/checkbox.modifier.ts
aabidsofi19: @nebula-aac  @sudhanshutech    we dont need to have a complex naming convention , lets agree of something simple and stay consistent

PR #470, opened 2024-01-27
Add and Refactor theme components

sudhanshutech:
**Notes for Reviewers**

This PR fixes #


packages/components/src/theme/components/button.modifier.ts
```diff
@@ -1,28 +1,34 @@
...
-    },
-    outlined: {
-      '&:hover': {
-        backgroundColor: CARIBBEAN_GREEN,
-        color: WHITE
-      }
+    root: ({ theme }) => {
+      const {
+        palette: {
+          secondary: { main }
```
sudhanshutech, line 9: @nebula-aac require your review here while defining secondary.
In `paletteOptions interface` primary was only defined , so do we required to define secondary there too? Or it will work?

sudhanshutech: And defining like this will also take care for ui which have only light mode like cloud? @nebula-aac

PR #621, opened 2024-05-17
feat(components): feedback default colour for both mode

captain-Akshay:
**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/src/custom/Feedback/style.tsx
+++ b/src/custom/Feedback/style.tsx
@@ -3,7 +3,6 @@ import {} from '../../constants/constants';
 import {
   BLACK,
   BUTTON_MODAL,
-  BUTTON_MODAL_DARK,
   CARIBBEAN_GREEN,
   CHINESE_SILVER,
   CULTURED,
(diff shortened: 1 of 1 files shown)


src/custom/Feedback/style.tsx
```diff
@@ -151,7 +150,7 @@ export const FeedbackSubmitButton = styled(Button)<SubmitProp>(({ isOpen }) => (
 }));
 
 export const FeedbackButton = styled(Button)<RenderPositionType>(({ theme, renderPosition }) => ({
-  backgroundColor: theme.palette.mode === 'dark' ? BUTTON_MODAL_DARK : BUTTON_MODAL,
+  backgroundColor: BUTTON_MODAL,
```
leecalcote, line 153: @captain-Akshay, is there any merit to leaving the support for dark mode and light mode in here, but simply having the same color for both?

leecalcote: @nebula-aac, I assume that, generally speaking, we prefer to have all uses of colors instrumented for light and dark mode support even if there is no change - just a general principle to follow.

captain-Akshay: since it was not visible with dark theme in meshmap with the side bar like we discussed. @leecalcote 
and the modal header is fixed already

leecalcote (approved):

PR #631, opened 2024-05-29
[Tooltip] Update zindex and title display condition

dragon-slayer875:
**Notes for Reviewers**

- In meshery UI, ziCalc is used to give tooltips a zindex of ziCalc(5), so I instead migrated it here, thus eliminating the need for ziCalc in meshery and meshery cloud.
Moreover, in meshmap  many components are given a high zindex which results in the tooltip not displaying properly.
In meshmap, no ziCalc value exceeds ziCalc(9) and thus I have set this value to remedy the problem there too.

- The tooltip was also not rendering content which was not in string format making it hard to catch bugs and refactor.
So I updated the condition for title usage.

### DIFF
--- a/src/custom/CustomTooltip/customTooltip.tsx
+++ b/src/custom/CustomTooltip/customTooltip.tsx
@@ -39,11 +39,12 @@ function CustomTooltip({
         },
         popper: {
           sx: {
+            zIndex: 99999,
             opacity: '1'
           }
         }
       }}
-      title={<RenderMarkdownTooltip content={typeof title === 'string' ? title : ''} />}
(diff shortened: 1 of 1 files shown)


src/custom/CustomTooltip/customTooltip.tsx
```diff
@@ -39,11 +39,12 @@ function CustomTooltip({
         },
         popper: {
           sx: {
+            zIndex: 9999999999,
```
aabidsofi19, line 42: @dragon-slayer875  why this large of zindex , make zIndex a prop .

aabidsofi19 (changes requested):

PR #642, opened 2024-06-05
fix margin for some tags in markdown support

sudhanshutech:
**Notes for Reviewers**

This PR fixes #


sudhanshutech:
@aabidsofi19 there is remark-breaks package also that integrate with markdown and it fix spacing issues, Should i include that?

PR #644, opened 2024-06-07
Fix click event for tooltip

sudhanshutech:
**Notes for Reviewers**

This PR fixes the click event on links inside tooltips which was mainly occurring inside modal . Moving `event propogation` below now ensures that the link is opened first, and then the event propagation is stopped.

Adding `as="a"` ensures that the StyledMarkdown component is rendered as an <a> tag, making it behave correctly as a hyperlink.

https://github.com/layer5io/sistent/assets/90546692/f0de8ad2-2fae-4315-8ba6-273757f74913


dragon-slayer875 (approved):
LGTM. Thank you. :smile:

PR #645, opened 2024-06-08
Add theming to toolbar options

sudhanshutech:
**Notes for Reviewers**

This PR adds theme to toolbar components which were missing.
image
image


Rexford74:
This looks largely good. There are two things I'd like to point out though.
First, what text token are you using here.
Second, the unchecked state for the checkbox doesn't have a green outline on it. I am aware that this is the initial version but after the last meeting that I attended on Friday, I made changes to the checkbox in Figma to include the green outline. I wonder if there has been a reconsideration on this stance...

leecalcote:
@sudhanshutech thoughts on @Rexford74's feedback?

sudhanshutech:
@leecalcote i tried the requested design of checkbox , here is the checbox screenshot. 
image

What i discovered from this is we should stick with our existing checkbox which is single border in a unchecked state.
**Reason:** is the checkbox appears quite small with the double outline, making it challenging for users to easily distinguish the clickable area specially in mobile screen it will hard for user to distinguish these two borders and will look might be confusing and cluttered.
`The existing with single outline gives a breathing space. `These were my findings and However, if needed, I can implement the style.

leecalcote:
> @leecalcote i tried the requested design of checkbox , here is the checbox screenshot. image
> 
> What i discovered from this is we should stick with our existing checkbox which is single border in a unchecked state. **Reason:** is the checkbox appears quite small with the double outline, making it challenging for users to easily distinguish the clickable area specially in mobile screen it will hard for user to distinguish these two borders and will look might be confusing and cluttered. `The existing with single outline gives a breathing space. `These were my findings and However, if needed, I can implement the style.
> 
> Let me know your thoughts // @Rexford74

@sudhanshutech demo this issue today, please.

sudhanshutech:
Ah sorry i missed, I will do surely next time.

captain-Akshay (approved):

PR #728, opened 2024-09-04
Update publish modal schema

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

This PR adds missing parameter in publish modal RJSF schema which tell RJSF to decode input data in URI encoded format and show data in input field in decoded format


Yashsharma1911:
From todays call we had a little confusion if we should accept this change or not, well yes we should accept because, Uzairs proposal was to migrate these schemas to schemas repo which was not related to fixing issue, and my proposal was to fix issue first because we previously used to have these schemas in schemas repo and then migrated to Sistent, now migrating them to again in schemas repo will require more work then which is needed right now

For now (right now) I'm focusing on fixing issue not creating extra work which I'm not sure should be a priority, I understand it is nice to have these schemas in schemas repo however I'm not sure if we should put this on priority right now, if we should do then let me know

leecalcote:
@MUzairS15 @aabidsofi19 thoughts?

src/schemas/publishCatalogItem/schema.tsx
```diff
@@ -24,14 +24,16 @@ const publishCatalogItemSchema = {
...
+      "x-encode-in-uri": true
     },
     pattern_info: {
       type: 'string',
       title: 'Description',
       description: 'Purpose of the design along with its intended and unintended uses.',
       format: 'textarea',
-      'x-rjsf-grid-area': 12
+      'x-rjsf-grid-area': 12,
+      "x-encode-in-uri": true
```
aabidsofi19, line 36: @Yashsharma1911  these are native to rjsf right ? . if not where are we documenting these . we should be using a uiSchema and not putting the presentation specific properties in the data structure schema

aabidsofi19 (changes requested):

aabidsofi19:
i think the publish schema should be derived from the schema for input of the publish endpoint .

leecalcote:
A relevant discussion - https://github.com/meshery/meshkit/pull/584#issuecomment-2322800552

src/schemas/publishCatalogItem/schema.tsx
Yashsharma1911: I agree @aabidsofi19 we can move it to uiSchema, we are tracking them here - https://github.com/layer5io/sistent/blob/master/src/schemas/readme.md#custom-properties

Yashsharma1911: Also in RJSF section in Meshmap developer guide

Yashsharma1911: Ops sorry, I mean "Kanvas"

Yashsharma1911:
> i think the publish schema should be derived from the schema for input of the publish endpoint . to remove the redundancy of redefining the same properties . for the presentation we should use a separate uiSchema ( that defines ui like the widget to use, grid size etc not the data structure )
> 
> // @leecalcote @Yashsharma1911 @MUzairS15

This makes much more sense, we can update the capability of uiSchema to support this, and one part is to move them in single place, it does make sense for me if these schemas get bundled in schema repo npm package.

Yashsharma1911:
There is issue with prettier, lemme fix it

leecalcote:
@Yashsharma1911 are we reordering fields, too?

Yashsharma1911:
I need approval to merge this PR, if looks good would be nice 

> @Yashsharma1911 are we reordering fields, too?

Ops I missed this, lemme fix it quick

leecalcote (approved):
LGTM

vishalvivekm (approved):

Yashsharma1911:
Thankyou for approving and merging my PR @vishalvivekm

PR #744, opened 2024-10-02
test(playwright): adding data-testid to support e2e tests

yash37158:
**Notes for Reviewers**
Added data-testid for supporting playwright e2e test for UI/Environments
This PR fixes [UI/environment Playwright test](https://github.com/meshery/meshery/pull/12088)


src/custom/TransferModal/TransferList/TransferList.tsx
```diff
@@ -215,7 +215,7 @@ function TransferList({
     listId?: string
   ) => (
     <StyledPaper>
-      <List dense component="div" role="list">
```
jerensl, line 218: I think it's better to use test id in meshery/meshery as the parent of this sistent element, and find them by role or aria-label. It's enough to distinguish the element i think. Wdyt?

https://playwright.dev/docs/api/class-framelocator#frame-locator-get-by-label

jerensl (changes requested):

jerensl:
> @yash37158 Are you suggesting to add `data-testid` to support `e2e` in `sistent` or from Meshery UI?
> 
> If it is meant for Meshery UI, I would suggest against this, as it can confuse future `e2e` testing for `sistent` versus Meshery UI.

I agree with this especially MUI styled do not accept non standard attributes like data-testid if im not mistaken

yash37158:
@jerensl I understand that the Meshery UI has some reliability issues when it comes to consistently loading components. Additionally, dependency injection isn't feasible within the Playwright tests. What other options are available?

PR #753, opened 2024-10-09
Convert cloud catalog sort menu and search field to sistent

amitamrutiya:
**Notes for Reviewers**

image

- [X] Yes, I signed my commits.


amitamrutiya:
@aabidsofi19 @sudhanshutech Now this catalog search and sort is useful for the open catalog model functionality. could you please review?

amitamrutiya:
Done @aabidsofi19

aabidsofi19:
confilcts here ,  make it a bit more generic ( not specific to catolog )

amitamrutiya:
@aabidsofi19 There is no need to create a sort by component for the catalog we can build them from the other sistent component. For the search one we already have `stylesearchbar` that we are not using anywhere so i added some styles to it so it looks like a cloud one so we can use it in other places.  

I hope it make sense to you.

PR #759, opened 2024-10-15
fix: updated docs

ianrwhitney:
**Notes for Reviewers**

This PR fixes #
This PR helps clarify how to attach and run sistent repo in another project. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@ build:
 
 ```
 attach sistent to your project:
-	npm install [path to sistent repo ]
+  npm pack (in sistent repo)
+	npm install /path/to/file.tgz (in another repo)
 ```
 
 > [!NOTE]


README.md
```diff
@@ -67,7 +67,8 @@ build:
 
 ```
 attach sistent to your project:
-	npm install [path to sistent repo ]
+  npm pack (in sistent repo)
+	npm install /path/to/file.tgz (in another repo)
```
sudhanshutech, line 71: why you changing this path to specific ?

ianrwhitney: i wasnt able to get sistent to attach when running meshery with the given commands in the readme. I had to pack in a tgz file and then ask npm to install that. This was the only way it was able to attach. If the orginial works for you maybe i can add this as an alternative here? Let me know your thoughts. thanks for your time to comment.

leecalcote: @nebula-aac, what do you think?

leecalcote: @SAHU-01 you might have a perspective to offer.

SAHU-01: @ianrwhitney if you wish to attach `sistent` with **meshery**, when you go into the `ui` folder of **meshery** and do a `npm i [sistent path]` it adds the `sistent` changes . If you attempt to attach `sistent` by executing `npm i [sistent path]` directly in `meshery` remote repo, it most likely doesn't work (at least in my case).

For the other repositories that I've added sistent code to, the `npm i [sistent path]` works well without an issue. So I believe we should keep the original one and maybe add a special case for meshery repo below to either run in inside ui or to use a specific path like your suggestion.

ianrwhitney: yea i can add this as an alternative. thanks for the comments

amitamrutiya:
Fixed this issue in the current README.

PR #846, opened 2024-12-11
Fix ResponsiveDataTable bug

### LINKED ISSUE #635: `ResponsiveDataTable` bug and readme updates
#### Current Behavior
- This minimal example, does'nt render the table and columns prop is ignored.
```tsx
import { ResponsiveDataTable } from "@layer5/sistent";

function App() {
  const columns = [
    { name: "name", label: "Name" },
    { name: "company", label: "Company" },
    { name: "city", label: "City" },
    { name: "state", label: "State" },
  ];

  const data = [
    ["Joe James", "Test Corp", "Yonkers", "NY"],
    ["John Walsh", "Test Corp", "Hartford", "CT"],
    ["Bob Herm", "Test Corp", "Tampa", "FL"],
    ["James Houston", "Test Corp", "Dallas", "TX"],
  ];

  const colVis = { name: true, company: false, city: true, state: true };

  return (
    <>
      <ResponsiveDataTable
        data={data}
        columns={columns}
        columnVisibility={colVis}
      />
    </>
  );
}

export default App;

```
image
- From the [implementation](https://github.com/layer5io/sistent/blob/44eb9698d484b4904e0ecc0e0056103cd8bcf0a3/src/custom/ResponsiveDataTable.tsx#L29) `tableCols` prop seems to be required to render the Table.
- Even with `tableCols` prop, `columnVisibility` seems to be ignored.
```tsx
import { ResponsiveDataTable } from "@layer5/sistent";

function App() {
  const columns = [
    { name: "name", label: "Name" },
    { name: "company", label: "Company" },
    { name: "city", label: "City" },
    { name: "state", label: "State" },
  ];

  const data = [
    ["Joe James", "Test Corp", "Yonkers", "NY"],
    ["John Walsh", "Test Corp", "Hartford", "CT"],
    ["Bob Herm", "Test Corp", "Tampa", "FL"],
    ["James Houston", "Test Corp", "Dallas", "TX"],
  ];

  const colVis = { name: true, company: false, city: true, state: true };

  return (
    <>
      <ResponsiveDataTable
        data={data}
        tableCols={columns}  //should be optional
        columns={columns}
        columnVisibility={colVis}
      />
    </>
  );
}

export default App;

```
image

#### Expected Behavior

- For each of these parameters - `columns`, `tableCols`, `columnVisibility`, Correct the implementation/readme to reflect correct behaviour.

#### Environment

- **Host OS:** Mac Linux Windows
- **Platform:** Docker or Kubernetes
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---

- 🎨 Wireframes and [designs for Sistent site](https://www.figma.com/file/ekm0Hh6M25bSKpP5BcIJom/Sistent-Design-System?type=design&node-id=596%3A5227&mode=design&t=8MExPykHN1vQFCTZ-1) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)

mdkaifansari04:
**Notes for Reviewers**
This PR fixes the bug if `ResponsiveDataTable` it has a required prop as `columns` and option props as `tableCols`( both the prop type is same) which is displayed in MUI datatable, if someone doesn't pass `tableCols` the table breaks.

So I fixed this by passing the both `columns` or `tableCols`, if the optional prop is not passed it wont break anymore. Altough the any big props changes may lead to make the chages all over where it is been used.

This PR fixes #635

### DIFF
--- a/src/custom/CatalogDetail/ActionButton.tsx
+++ b/src/custom/CatalogDetail/ActionButton.tsx
@@ -4,7 +4,7 @@ import { CopyIcon, EditIcon, KanvasIcon, PublishIcon } from '../../icons';
 import Download from '../../icons/Download/Download';
 import { charcoal, useTheme } from '../../theme';
 import { Pattern } from '../CustomCatalog/CustomCard';
-import { downloadFilter, downloadYaml } from './helper';
+import { downloadPattern, downloadYaml } from './helper';
 import { ActionButton, StyledActionWrapper, UnpublishAction } from './style';
 import { RESOURCE_TYPES } from './types';
 
@@ -79,9 +79,9 @@ const ActionButtons: React.FC<ActionButtonsProps> = ({
               color: theme.palette.text.default
             }}
             onClick={() =>
-              cleanedType === RESOURCE_TYPES.FILTERS
-                ? downloadFilter(details.id, details.name)
-                : downloadYaml(details.pattern_file, details.name)
+              cleanedType === RESOURCE_TYPES.VIEWS
+                ? downloadYaml(details.pattern_file, details.name)
+                : downloadPattern(details.id, details.name, cleanedType)
             }
           >
             <Download width={24} height={24} fill={theme.palette.icon.default} />
--- a/src/custom/CatalogDetail/helper.ts
+++ b/src/custom/CatalogDetail/helper.ts
@@ -34,8 +34,9 @@ export function slugify(str: string): string {
   return str;
 }
 
-export const downloadFilter = (id: string, name: string): void => {
-  const dataUri = `${process.env.API_ENDPOINT_PREFIX}/api/content/filters/download/${id}`;
+export const downloadPattern = (id: string, name: string, type: string): void => {
+  const pattern = type == 'design' ? 'patterns' : 'filters';
+  const dataUri = `${process.env.API_ENDPOINT_PREFIX}/api/content/${pattern}/download/${id}`;
 
   // Add the .wasm extension to the filename
   const fileNameWithExtension = name + '.wasm';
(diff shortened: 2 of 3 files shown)


capricorn-32:
The MUIDatatable [documentation](https://github.com/gregnb/mui-datatables?tab=readme-ov-file#muidatatable-)  specifies that there is no `tableCols` prop. However, in the codebase, there are inconsistencies: some components pass both `columns` and `tableCols` with same arguments, while others use only `columns`.

To eliminate this ambiguity, it’s important to note that `columns` is designed to define the table headers and manage the visibility of specific columns. Standardizing the use of `columns` across the codebase would avoid confusion.

mdkaifansari04:
You are completly right @capricorn-32 but sistent compoent are the base component whihc is been used across all the projects of Layer5, making any major change in the prop type or removing any prop, will lead to a mojor break and leads to breaking change across all over the places it is been used.

Although all the components is been documneted. 

Just to avoid all this pain, I tried to make a change in a way a that, it should be fixed without changing the behavoiur or the current component. 

I can remove the `tableCols` props as there is no use of it, but it will lead us to major change in all over the code bases.

capricorn-32:
@mdkaifansari04 I am pointing this because in future it will more pain to work around this. Lets see, what decision should be made to prevent this.

mdkaifansari04:
One way to fix this would be change in component and update the docs,

After just the release we have to make changes in the corresponding projects as I've access to all the projects eventually. 

I can make the changes but I'm busy in some school stuff might not be possible now. 

If anyone shows up we can create issues for each repo.

sudhanshutech:
@mdkaifansari04 thanks for looking into the issue but i guess tablecols is not a optional column but a required one because the state updatecols is updating the columns visibility when we uncheck any column to view/hide. And also this pr includes few unnecessary changes.

PR #873, opened 2025-01-09
(ci): Merge Lint Check Workflows into a Single Workflow `checks.yml`

### LINKED ISSUE #783: [CI] Merge two separate workflows that does `lint` check
#### Current Behavior

There are two workflows:

- `Link check`
- `Checks`

Both of the workflow already includes a check to do a `lint` check.  There is no need for two separate workflows that does lint checks.

#### Acceptance Tests

---

- 🎨 Wireframes and [designs for Sistent site](https://www.figma.com/file/ekm0Hh6M25bSKpP5BcIJom/Sistent-Design-System?type=design&node-id=596%3A5227&mode=design&t=8MExPykHN1vQFCTZ-1) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)

usamaidrsk:
**Notes for Reviewers**

This PR fixes #783


.github/workflows/checks.yml
```diff
@@ -1,35 +1,41 @@
 name: Checks
```
sudhanshutech, line 1: @usamaidrsk will you rename this workflow to meaningful title please. `Checks` doesn't give idea to contributors what exactly the check is about.

sudhanshutech (commented):
@nebula-aac do we need to test for each nodes 16, 18, 20. And if we should then could it be better if those three can also be checked at once and produce logs when either of it fails. Let us know you thoughts?

.github/workflows/node-checks.yml
```diff
@@ -0,0 +1,44 @@
+name: Node.js Compatibility Checks
```
sudhanshutech, line 1:
```suggestion
name: Node version and Lint Check
```

sudhanshutech (changes requested):

PR #905, opened 2025-02-03
feat: add new electedComponents token to sistent with new grey shades

amitamrutiya:
**Notes for Reviewers**

This PR fixes #


src/theme/palette.ts
```diff
@@ -19,6 +19,7 @@ declare module '@mui/material/styles' {
       default: string;
     };
     tabs?: string;
+    elevatedComponents?: string;
```
sudhanshutech, line 22: @amitamrutiya we need to sure of this.. we don't want to add more and more tokens based on our needs otherwise we will be doing same mistake as it was in meshery of having so many colors. And we should use more genric names for such.

PR #906, opened 2025-02-03
fix: styling in panel component

Vidit-Kushwaha:
**Notes for Reviewers**

This PR fixes styling issues in panel component in Sistent library.

### DIFF
--- a/src/custom/Panel/style.tsx
+++ b/src/custom/Panel/style.tsx
@@ -1,7 +1,7 @@
 import { ListItemProps } from '@mui/material';
 import { Box, IconButton, ListItem } from '../../base';
 import { PanelDragHandleIcon } from '../../icons/PanelDragHandle';
-import { styled } from '../../theme';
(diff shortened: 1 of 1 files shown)


src/custom/Panel/style.tsx
```diff
@@ -88,19 +89,20 @@ export const PanelContainer = styled(Box)<{ intitialPosition: PanelProps['intiti
...
     position: 'absolute',
     backgroundColor: theme.palette.background.blur?.light,
-    boxShadow: '0 4px 16px #05003812',
+    boxShadow: '0 4px 16px black',
     maxHeight: '80%',
     display: 'flex',
     boxSizing: 'border-box',
     ...intitialPosition
   })
 );
```
amitamrutiya, line 99: don't use the hardcode color use token if the token is not present and you want to show the same color in both light and dark mode than use it from the colors.ts file

amitamrutiya (changes requested):

PR #907, opened 2025-02-06
chore: added fontfamily for base component

Vidit-Kushwaha:
**Notes for Reviewers**

This PR partially fixes [#12739](https://github.com/meshery/meshery/issues/12739) Consistent Qanelas Soft font family

**[Signed commits](../blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**

- [x] Yes, I signed my commits.

PR #914, opened 2025-02-12
Revert "chore: added fontfamily for base component"

amitamrutiya:
Reverts layer5io/sistent#907

@Vidit-Kushwaha


Vidit-Kushwaha:
Revert changes make sense for now. I will look in depth for other components.

PR #923, opened 2025-02-18
Add missing components that are present in meshery ui and extensions

amitamrutiya:
**Added Components**
  - SvgIcon
  - Zoom
  - Step
  - StepButton
  - StepConnector
  - StepContent
  - StepIcon
  - StepLabel
  - Stepper
  - Slider
  - Snackbar
  - SpeedDial
  - NativeSelect
  - Radio
  - InputBase
  - LinearProgress
  - Hidden
  - Fab
  - Fade
  - Autocomplete
  - Alert
  - TableBody
  - TableCell
  - TableContainer
  - TableHead
  - TableRow


package.json
```diff
@@ -110,6 +110,27 @@
     },
     "react-dom": {
       "optional": true
+    },
```
leecalcote, line 113: Cool. I just learned that sometimes, a package might have peer dependencies that are not strictly required for its core functionality. In such cases, you can use peerDependenciesMeta to mark a peer dependency as optional.

leecalcote (approved):

package.json
amitamrutiya: The list is based on what are the libraries that we are expected to already present in the project. I removed some libraries from the peer dependencies because it doesn't make sense that all other project like layer5io website are using them.

PR #1053, opened 2025-05-25
UI: Enhance Error Message Specificity in User Invite Modal

### LINKED ISSUE #1050: [Feature] Improve Error Reporting for Failed User Invitations
#### Current Behavior

This issue addresses the need for more specific error feedback when user invitations fail.

Currently, all user invitation failures result in a generic error message:
Invitation to ${inviteeFirstName} ${inviteeLastName} failed.

This message does not provide insight into the specific reason for the failure.

#### Desired Behavior

The system should return distinct error messages based on the actual cause of the invitation failure. These specific error messages should be derived from the error handling logic within `cloud/handlers/users.go`.

For example, instead of a generic failure message, users should see messages like:

"Invitation failed: Email address already exists."
"Invitation failed: Invalid email format."
"Invitation failed: User limit reached for the organization." (These are illustrative examples; actual messages will depend on the specific error conditions in cloud/handlers/users.go)

zihanKuang:
**Notes for Reviewers**

This PR fixes #1050

### Reason for Changes

The primary goal is to improve the user experience by providing more specific and actionable feedback when an invitation fails due to common, understandable reasons. This allows users to better diagnose the issue.

By whitelisting specific messages for detailed display, I avoid exposing end-users to potentially confusing or overly technical error messages.

image

### DIFF
--- a/src/custom/DashboardWidgets/GettingStartedWidget/InviteUserModal.tsx
+++ b/src/custom/DashboardWidgets/GettingStartedWidget/InviteUserModal.tsx
@@ -205,9 +205,51 @@ export default function UserInviteModal({
       }).unwrap();
 
       handleSuccess(`Invite send to ${inviteeName.trim() === '' ? inviteeEmail : inviteeName}.`);
-    } catch (e) {
-      console.debug('cannot send user invite', e);
-      handleError(`Invitation to ${inviteeFirstName} ${inviteeLastName} failed.`);
+    } catch (errorFromInvite: any) {
+      console.debug('Cannot send user invite, API error:', errorFromInvite);
+
+      const inviteeNameForFallback =
+        (inviteeFirstName + ' ' + inviteeLastName).trim() || inviteeEmail;
+      let displayErrorMessage = `Invitation to ${inviteeNameForFallback} failed. Please try again.`;
+
+      const whitelistedErrorMessages = [
+        'invalid org id',
+        'failed to retrieve organization',
+        'failed to add user to organization',
+        'failed to add user to team',
+        'user invite failed'
+      ];
+
+      let backendMessage: string | null = null;
+      if (errorFromInvite?.data) {
+        if (typeof errorFromInvite.data === 'string' && errorFromInvite.data.trim()) {
+          backendMessage = errorFromInvite.data.trim();
+        } else if (
+          errorFromInvite.data.message &&
+          typeof errorFromInvite.data.message === 'string' &&
+          errorFromInvite.data.message.trim()
+        ) {
+          backendMessage = errorFromInvite.data.message.trim();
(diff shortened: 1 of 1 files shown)


src/custom/DashboardWidgets/GettingStartedWidget/InviteUserModal.tsx
```diff
@@ -205,50 +207,33 @@ export default function UserInviteModal({
...
-        ) {
-          backendMessage = errorFromInvite.data.error.trim();
-        }
+    } catch (error: any) {
+      console.debug('cannot send user invite, API error:', error);
+      let displayErrorMessage = `Invitation to ${inviteeName.trim() === '' ? inviteeEmail : inviteeName} failed.`;
+      if (
+        error &&
+        error.data &&
+        typeof error.data === 'object' &&
```
leecalcote, line 216: Create new error codes in the respective handlers.

leecalcote (changes requested):

FaheemOnHub:
> Hi @FaheemOnHub,
> 
> I've just pushed the latest updates to my backend PR: [layer5io/meshery-cloud#3527](https://github.com/layer5io/meshery-cloud/pull/3527). Could you please take a look and let me know if any further modifications are needed?
> 
> Separately, regarding my frontend PR, could you advise if this should now be closed, or if it still requires further updates or work from my side?😊
> 
> Thanks!

as currently we don't have notification center, we could just display the short description in err.

src/custom/DashboardWidgets/GettingStartedWidget/InviteUserModal.tsx
```diff
@@ -205,9 +207,34 @@ export default function UserInviteModal({
...
-      console.debug('cannot send user invite', e);
-      handleError(`Invitation to ${inviteeFirstName} ${inviteeLastName} failed.`);
+    } catch (error: any) {
+      console.debug('cannot send user invite, API error:', error);
+      let displayErrorMessage = `Invitation to ${inviteeName.trim() === '' ? inviteeEmail : inviteeName} failed.`;
+      if (
+        error &&
+        error.data &&
+        typeof error.data === 'object' &&
+        error.data.code === INVITE_EXISTING_USER_ERROR_CODE
```
FaheemOnHub, line 217: is it necessary to hardcode, we can directly display the error message 🤔

zihanKuang:
Hi @FaheemOnHub, thanks for the feedback!

You're right, but the backend can send various error messages. My thinking was to specifically check for `error.data.code === INVITE_EXISTING_USER_ERROR_CODE` so we can provide user-friendly guidance for that particular scenario. For other potential errors from this endpoint, I fall back to a more general message. So I use hardcode to identify.

PR #1067, opened 2025-06-03
fix: view-modal fails to hold multiple entries at once across whole cloud ui.

FaheemOnHub:
This PR fixes https://github.com/layer5io/meshery-cloud/issues/3567

### DIFF
--- a/src/custom/UserSearchField/UserSearchFieldInput.tsx
+++ b/src/custom/UserSearchField/UserSearchFieldInput.tsx
@@ -19,7 +19,7 @@ import { iconSmall } from '../../constants/iconsSizes';
 import { CloseIcon, PersonIcon } from '../../icons';
 
 interface User {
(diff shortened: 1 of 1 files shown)


aabidsofi19:
@FaheemOnHub  why are we removing user.id it should works also for identifying or filtering out users

leecalcote:
@FaheemOnHub what are your thoughts on this one?

FaheemOnHub:
> @FaheemOnHub why are we removing user.id it should works also for identifying or filtering out users

We are not removing user.id , we are adding user.id 
basically we were using .id to search , which does not exist in usersData (was wrong)
Screenshot 2025-06-19 at 8 54 23 PM

so because of that it was taking all users as one , and when one was added , it shows(no user found).

FaheemOnHub:
we can merge this!!!

PR #1081, opened 2025-06-17
fix: Unify version display for public and private designs

zihanKuang:
**Notes for Reviewers**

This PR fixes: https://github.com/layer5io/meshery-cloud/issues/3585

#### Reason

There are three relevant states for a design: **private**, **public**, and **published**. The `private` and `public` states should both reflect the current draft's version number, which is stored in the `version` field. Only a `published` design should display the `published_version`.

The bug was that the UI was incorrectly displaying the `published_version` for designs that were set to `public`. This created problems:
1.  **Inconsistent Versions**: If a public design had no `published_version`, the UI would show an incorrect fallback value (like `0.0.0` or `""`).
2.  **Wrong Versions**: If a public design has a `published_version`, the UI would show an incorrect value.

#### Fix

https://github.com/user-attachments/assets/8ad17238-9beb-43e3-84fc-cf62246fba2a

### DIFF
--- a/src/custom/CustomCatalog/Helper.ts
+++ b/src/custom/CustomCatalog/Helper.ts
@@ -61,7 +61,7 @@ export const handleImage = async ({
 export const DEFAULT_DESIGN_VERSION = '0.0.0';
 
 export const getVersion = (design: Pattern) => {
-  if (design.visibility === 'public') {
+  if (design.visibility === 'published') {
     return design?.catalog_data?.published_version || DEFAULT_DESIGN_VERSION;
   }
   try {
--- a/src/custom/CustomCatalog/Helper.ts
+++ b/src/custom/CustomCatalog/Helper.ts
@@ -42,6 +42,18 @@ const getValidSvgPaths = async (
   return validSvgPaths;
 };
 
+// Parses the pattern file content (YAML) and returns the version field.
+// If parsing fails or version is missing, returns the default version.
+const getWorkingVersionFromPatternFile = (patternFileContent: string) => {
+  try {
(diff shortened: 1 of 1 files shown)


src/custom/CustomCatalog/Helper.ts
```diff
@@ -61,7 +61,7 @@ export const handleImage = async ({
 export const DEFAULT_DESIGN_VERSION = '0.0.0';
 
 export const getVersion = (design: Pattern) => {
-  if (design.visibility === 'public') {
+  if (design.visibility === 'published') {
```
leecalcote: In the event that the design is "public", what happens? @zihanKuang

leecalcote: Where is "private" handled?

zihanKuang: Previously, if the design was `public`, it would display the `published_version`. After the changes, only `published` designs will show `published_version`. `Public` designs will now display their regular version

For `private` designs, they were already displaying the correct `version`. With the new logic, they now fall through the `if` condition and also display their `version` from within the `try` block.

leecalcote: 1. Perhaps, use of a `case` / `switch` statement is more appropriate and easier to understand/maintain. Designing for intentional failure isn't quite what we should do, I believe. Instead, the `try` block or an `else` block should focus on error handling as opposed to handling the core use case that will be invoked the majority of the time.
2. Add comments to the code to make the handling of versions by visibility level easier to understand, if you would, please.

PR #1117, opened 2025-07-30
Migration of own-learning card from sistent to cloud-ui

FaheemOnHub:
**Notes for Reviewers**

Initially the ownLearningcard was placed in sistent as it was not much different from other learning cards but now it includes a modal and follows a stepper component which does not makes it suitable to place in sistent.

### DIFF
--- a/src/custom/LearningCard/LearningCard.tsx
+++ b/src/custom/LearningCard/LearningCard.tsx
@@ -1,8 +1,5 @@
-import React, { useState } from 'react';
-import { Grid2, Typography } from '../../base';
-import { ExternalLinkIcon } from '../../icons';
-import { Modal, ModalBody, ModalButtonPrimary, ModalButtonSecondary, ModalFooter } from '../Modal';
-import { CopyToClipboard } from '../ResourceDetailFormatters/Component';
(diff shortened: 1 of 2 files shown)


ritzorama (approved):

src/custom/LearningCard/LearningCard.tsx
```diff
@@ -35,43 +31,17 @@ interface Props {
...
-      target={isExternal ? '_blank' : undefined}
-      rel={isExternal ? 'noopener noreferrer' : undefined}
-    >
-      {children}
-    </CardLink>
-  );
+  return <CardLink href={path}>{children}</CardLink>;
 };
 
-const LearningCard: React.FC<Props> = ({
```
aabidsofi19, line 62: lets rename this component to ContentOverviewCard , and the tutorial prop to curriculaContent and use the types from schemas .  couse count to unitCount  and word "courses" to units . the create new content needs to be separate content only reusing the base card styling and not a part of this component .

aabidsofi19: i guess these changes need to be done in cloud ui though

PR #1159, opened 2025-09-09
line height reduced of sistent button

saurabhraghuvanshii:
**Notes for Reviewers**
### before 
Screenshot From 2025-09-10 15-42-51

### after 
Screenshot From 2025-09-10 15-42-58

This PR fixes #

### DIFF
--- a/src/custom/Typography/index.tsx
+++ b/src/custom/Typography/index.tsx
@@ -71,6 +71,22 @@ export const TextB2SemiBold = styled(Typography)<{
   }
 }));
 
+export const TextB2Medium = styled(Typography)<{
+  mode?: 'light' | 'dark' | PaletteMode;
+}>(({ theme, mode = 'light' }) => ({
+  fontFamily: ['Qanelas Soft Regular'].join(','),
+  color: mode === 'light' ? common.black : common.white,
+  fontSize: '1rem',
+  fontWeight: 500,
+  lineHeight: '1.25rem',
+  textTransform: 'capitalize',
+  display: 'block',
+  [theme.breakpoints.down('sm')]: {
+    fontSize: '1rem',
+    lineHeight: '1.25rem'
+  }
+}));
+
 export const TextB3Regular = styled(Typography)(() => ({
   fontFamily: ['"Open Sans"', 'sans-serif'].join(','),
   fontSize: '0.875rem',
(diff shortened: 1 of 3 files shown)


src/theme/components/button.modifier.ts
```diff
@@ -14,6 +14,7 @@ export const MuiButton: Components<Theme>['MuiButton'] = {
       return {
         ...textB2SemiBold,
         fontWeight: 500,
+        lineHeight: '1.25rem',
```
leecalcote: Should this be a statically assigned line height?

What is the line height currently set to; currently inherited as?

saurabhraghuvanshii: Currently by default it is 1.75rem

sudhanshutech: @saurabhraghuvanshii this is i guess not the right way to directly update the lineheight. It should be done through typography here https://github.com/layer5io/sistent/blob/9a8d974835afffaf5fb7dda76384898a9f9d8375/src/theme/typography.ts.
Currently its used textB2semibold, so see if any other use case works better here with same font but different line height but it should in any case i guess affected through figma sistent documentation.

saurabhraghuvanshii: @sudhanshutech ok I'll look into this find better solution

saurabhraghuvanshii: @sudhanshutech how is that I create i new component in typography.ts like 
```
textB2Medium: {
      fontFamily: ['Qanelas Soft Regular'].join(','),
      color: mode === 'light' ? common.black : common.white,
      fontSize: '1rem',
      fontWeight: 500,
      lineHeight: '1.25rem',
      textTransform: 'capitalize',
      [theme.breakpoints.down('sm')]: {
        fontSize: '1rem',
        lineHeight: '1.25rem'
      }
    },
```

which is replace current `textB2SemiBold`

saurabhraghuvanshii: ### before 
Screenshot From 2025-09-10 15-42-51

### after
 
Screenshot From 2025-09-10 15-42-58

src/custom/Typography/index.tsx
```diff
@@ -71,6 +71,22 @@ export const TextB2SemiBold = styled(Typography)<{
   }
 }));
 
+export const TextB2Medium = styled(Typography)<{
```
leecalcote, line 74: We're not looking for a new component, just a minor tweak on the existing Button.

leecalcote: It's a line-height PR... should a one-liner

Issue #1193, opened 2025-12-30
[Bug] Incorrect WarningIcon


Rajesh-Nagarajan-11:
#### Current Behavior

the WarningIcon currently used in the UI is incorrect and does not match the standard warning icon 

<img width="1362" height="611" alt="Image" src="https://github.com/user-attachments/assets/baf6500e-e8ec-4c50-938d-8ca094621956" />

#### Expected Behavior
Use the correct and standardized WarningIcon

<img width="284" height="290" alt="Image" src="https://github.com/user-attachments/assets/4cf1414a-1f03-4010-b9c1-c8e70e365e4d" />

---

#### Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources

- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/build-and-release)
- 📚 [Instructions for contributing to documentation](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#documentation-contribution-flow)
  - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Sistent site](https://www.figma.com/file/ekm0Hh6M25bSKpP5BcIJom/Sistent-Design-System?type=design&node-id=596%3A5227&mode=design&t=8MExPykHN1vQFCTZ-1) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

PR #1194, opened 2025-12-30
ISSUE #1193. Now show correct warning icon 

JFQR:
BEFORE: in layer5 page, when warning icon was searched, a wrong icon was showed (img1)
AFTER: changed the path of the warning icon to match a correct svg (img 2)
grafik
grafik


Rajesh-Nagarajan-11:
Issue is already Fixed !

PR #1199, opened 2026-01-07
[UI] Hide workspace actions for non-permitted users

Rajesh-Nagarajan-11:
**Notes for Reviewers**

Implemented conditional rendering so that only users with the required permissions can view and perform actions such as
- Bulk select checkbox
- Edit 
- Delete 

Removed line 360 ->> `disabled={deleted ? true : !isDeleteWorkspaceAllowed}` since disabling the checkbox after deleting the workspace is meaningless

Removed all disabled props from the actions since they are no longer needed.

**Output after connecting with local Sistent (current role: Org Admin)**

[Screencast from 2026-01-07 15-24-10.webm](https://github.com/user-attachments/assets/8d2c7db0-4498-41db-8636-a5a91c02a5ec)

**Output after connecting with local Sistent (current role: User)**

[Screencast from 2026-01-07 15-28-33.webm](https://github.com/user-attachments/assets/916761ae-f2cc-4b19-bc4a-0221219419d8)


Rajesh-Nagarajan-11:
@leecalcote To be successful in this PR, what am I supposed to do ?

PR #1274, opened 2026-01-30
feat: make notify checkbox optional

ritzorama:
- add showNotifyCheckbox prop (default true) to UserSearchField\n- render notify checkbox only when explicitly allowed and handler provided\n- bump version to 0.0.3

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sistent/sistent",
-  "version": "0.0.3",
+  "version": "0.16.5",
   "description": "Reusable React Components and SVG Icons library",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sistent/sistent",
-  "version": "0.0.2",
+  "version": "0.0.3",
```
fitzergerald: Don't be an idiot, @ritzorama.


```suggestion
  "version": "0.16.5",
```

fitzergerald (approved):

PR #1431, opened 2026-04-24
fix(schemas): Phase 2.K — align Sistent with canonical camelCase wire format

miacycle:
## Summary

Phase 2.K of the [identifier-naming migration](https://github.com/meshery/schemas/blob/master/docs/identifier-naming-migration.md) — aligning the Sistent library with the canonical camelCase wire format now published by the upstream Workspace / Environment / Model APIs.

After Phase 3 + Phase 2 tail PRs merged across `meshery/schemas`, `meshery/meshery`, and `layer5io/meshery-cloud`, the v1beta3 Workspace/Environment APIs and the v1beta2 Model API now return `organizationId` (and friends) on the wire. Sistent still pinned `@meshery/schemas@^1.0.5` and referenced `organization_id` in several places, which silently yields `undefined` against real responses — an empty "Organization ID" column in `EnvironmentTable` and broken form binding on the create/edit Workspace and Environment modals.

The earlier consumer-audit (`meshery/schemas#...`) did not catch this because its TypeScript parser scans `meshery/ui`, `meshery-cloud/ui`, and `meshery-extensions` only; the Sistent library is out of its reach.

### Upstream PRs this completes

- `meshery/schemas#830`, `#831` — v1beta3 Workspace + Environment canonical schemas, v1beta2 Model
- `layer5io/meshery-cloud#5092` — handler migration to camelCase wire
- `meshery/meshery#18904`, `#18905` — server + UI migration

### Dependency

- `@meshery/schemas`: `^1.0.5` → `^1.1.1` (picks up canonical v1beta3 Workspace/Environment + v1beta2 Model exports)

### Schema imports repointed

| File | Before | After |
|---|---|---|
| `src/schemas/createAndEditEnvironment/schema.tsx` | `EnvironmentDefinitionV1Beta1OpenApiSchema` | `EnvironmentDefinitionV1Beta3OpenApiSchema` (subpath import) |
| `src/schemas/createAndEditWorkspace/schema.tsx` | `WorkspaceDefinitionV1Beta1OpenApiSchema` | `WorkspaceDefinitionV1Beta3OpenApiSchema` (subpath import) |
| `src/schemas/importModel/schema.tsx` | `ModelDefinitionV1Beta1OpenApiSchema` | `ModelDefinitionV1Beta2OpenApiSchema` (subpath import) |

[...]

PR #1433, opened 2026-04-24
ci(release): make npm-version step idempotent to unblock v0.19.0 publish

leecalcote:
## Problem

The v0.19.0 release was cut on 2026-04-24 after \`layer5io/sistent#1431\` merged, but the Publish Node.js Package workflow [failed](https://github.com/layer5io/sistent/actions/runs/24892445156) at the \"Set Package Version\" step:

\`\`\`
Error: Command failed: npm version v0.19.0 --prefix ./ --no-git-tag-version
npm error Version not changed
\`\`\`

Root cause: the merged PR #1431 committed \`package.json#version = \"0.19.0\"\` so that master's metadata reflects the next published version. The release workflow's \`reedyuk/npm-version@1.1.1\` then tries to bump to the same version and \`npm version\` exits 1 because the target matches current. The result: the GitHub release v0.19.0 exists but \`@sistent/sistent@0.19.0\` was never published to npm — npm still shows 0.18.8 as latest.

## Fix

Replace the \`reedyuk/npm-version\` action with an inline bash step that calls \`npm version --allow-same-version\`. The step is now idempotent: if \`package.json\` already matches the tag, it's a no-op; otherwise it performs the bump. No other behavior changes.

\`\`\`yaml
- name: \"Set Package Version\"
  env:
    TAG_NAME: \${{ github.event.release.tag_name || inputs.tag_name }}
  run: |
    VERSION=\"\${TAG_NAME#v}\"
    npm version \"\$VERSION\" --no-git-tag-version --allow-same-version
\`\`\`

## Unblocks

After this merges, re-trigger v0.19.0 publish via \`workflow_dispatch\` (tag_name: v0.19.0) or by deleting + recreating the release. \`@sistent/sistent@0.19.0\` will land on npm, unblocking:
- meshery/ui Sistent pin bump PR
- meshery-cloud/ui Sistent pin bump PR

## Relationship to #1432

Complementary, not overlapping. PR #1432 adds a *commit-back* step *after* publish so master's \`package.json\` stays truthful post-release. This PR fixes the *bump* step itself so publish doesn't error when master is already truthful (which is the intended outcome once #1432 merges). Both PRs can land in either order.

## Test plan

- [x] YAML validates (\`python3 -c 'import yaml; yaml.safe_load(open(\".github/workflows/release.yml\"))'\`).
- [x] \`actionlint\` clean on the changed file (if run locally).
- [ ] After merge: \`workflow_dispatch\` on Publish Node.js Package with \`tag_name: v0.19.0\` succeeds end-to-end.
- [ ] \`npm view @sistent/sistent version\` shows \`0.19.0\`.

## Rollback

Revert this one commit to restore the \`reedyuk/npm-version@1.1.1\` step.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,9 +38,16 @@ jobs:
           scope: "@sistent"
 
       - name: "Set Package Version"
-        uses: reedyuk/npm-version@1.1.1
-        with:
-          version: ${{ github.event.release.tag_name || inputs.tag_name }}
+        env:
+          TAG_NAME: ${{ github.event.release.tag_name || inputs.tag_name }}
+        run: |
+          # Strip a leading 'v' from the release tag (v0.19.0 -> 0.19.0) and
+          # set package.json#version. --allow-same-version makes the step
+          # idempotent when master's package.json already matches the tag
+          # (e.g. the merged PR that cut the release had committed the bump).
+          # --no-git-tag-version prevents npm from creating an extra tag.
+          VERSION="${TAG_NAME#v}"
+          npm version "$VERSION" --no-git-tag-version --allow-same-version
 
       - name: Install, Build, and Publish Package
         run: |


miacycle (approved):
Minimal surgical hotfix. Idempotent npm version call matches the intent; rollback is trivial.

PR #1434, opened 2026-04-24
fix(schemas): align importDesign with oneOf[File|URL] shape (0.19.1)

miacycle:
## Summary

Fixes the Next.js SSR crash `TypeError: Cannot read properties of undefined (reading 'name')` that surfaced downstream in meshery-cloud staging after pinning `@sistent/sistent@0.19.0` (meshery-cloud PR [#5094](https://github.com/layer5io/meshery-cloud/pull/5094)) and that was also flagged on PRs #1431 / #1433 but deferred as "out of scope."

### Root cause

[meshery/schemas commit b08a4f62](https://github.com/meshery/schemas/commit/b08a4f62) tightened `MesheryPatternImportRequestBody` from a flat `properties` object to a `oneOf: [MesheryPatternImportFilePayload, MesheryPatternImportURLPayload]` union, matching the server handler's "exactly one of File Import or URL Import" contract. That contract shipped in `@meshery/schemas@1.1.0` and is re-exposed here through the bundled `DesignDefinitionV1Beta1OpenApiSchema`.

`src/schemas/importDesign/schema.tsx` still walked `.properties.name`, `.properties.file`, `.properties.url` at the root of the request body. With the `oneOf` shape those paths resolve to `undefined`, and module evaluation throws at the first `.properties.name.type` read — before any React component even mounts. Because Sistent bundles `@meshery/schemas` inline (`noExternal: [/^@meshery\\/schemas/]` in `tsup.config.ts`), every downstream Next.js SSR build ships the same crash at `Collecting page data`.

### Fix

Resolve the individual field shapes from the dedicated payload components so the UI schema continues to be driven by the API contract rather than duplicating it:

- `name` and `file` come from `MesheryPatternImportFilePayload.properties`
- `url` comes from `MesheryPatternImportURLPayload.properties`
- The top-level prompt description still sources from the wrapper `MesheryPatternImportRequestBody.description`

No behavior change for the rendered form — same fields, same descriptions, same validation.

### Version bump

Bumps to `0.19.1`. Downstream consumers pinned `^0.19.0` (meshery-cloud, meshery-ui, meshery-extensions) will pick this up on their next install without any consumer-side change.

## Test plan

- [x] `npm run build` succeeds (`tsup` produces `dist/index.js` and `dist/index.mjs`).
- [x] `npm test` passes (7 suites / 21 tests).
- [x] `require('@sistent/sistent')` no longer throws at module load.
- [x] `meshery-cloud/ui` `npm run build-staging` completes successfully with the patched sistent overlaid into `node_modules/` (previously crashed at `Collecting page data using 7 workers`).
- [x] `importDesignSchema` resolves to a well-formed JSON schema with `properties.name.title === 'Design file name'`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@sistent/sistent",
-  "version": "0.19.0",
+  "version": "0.19.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@sistent/sistent",
-      "version": "0.19.0",
+      "version": "0.19.1",
       "dependencies": {
         "@emotion/react": "^11.14.0",
         "@emotion/styled": "^11.14.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sistent/sistent",
-  "version": "0.19.0",
+  "version": "0.19.1",
   "description": "Reusable React Components and SVG Icons library",
   "repository": {
     "type": "git",
--- a/src/schemas/importDesign/schema.tsx
+++ b/src/schemas/importDesign/schema.tsx
@@ -3,15 +3,32 @@ import { DesignDefinitionV1Beta1OpenApiSchema } from '@meshery/schemas';
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const DesignSchema = (DesignDefinitionV1Beta1OpenApiSchema as any).components.schemas;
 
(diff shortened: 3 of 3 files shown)


leecalcote (approved):
Fixes the 0.19.0 SSR crash at module load (MesheryPatternImportRequestBody.properties undefined under the oneOf contract). Verified locally by overlaying the rebuilt tarball into meshery-cloud/ui and running the staging build end-to-end — passes Collecting page data and onward. Merging to cut v0.19.1.

PR #1477, opened 2026-05-04
fix(CatalogDesignTable): pass design ID as third argument to handleCopyUrl for backward compatibility

rishiraj38:
## Description

Fixes the "Copy Link" action in the catalog table view generating a URL with `catalog-design=undefined` instead of the actual design ID.

### Root Cause

The `handleCopyUrl` callback in `createDesignColumns` was called with two arguments:
`handleCopyUrl(rowData.id, rowData.name)`

However, the extension consumer's handler expects the design ID in the **third** parameter position (a legacy signature from when the call included a source type as the first argument). Since only two arguments were passed, the third parameter evaluated to `undefined`, resulting in:
`https://playground.meshery.io/extension/meshmap?catalog-design=undefined`

### Fix

Passed `rowData.id` in both the first and third argument positions to maintain backward compatibility with the extension handler:
`handleCopyUrl(rowData.id, rowData.name, rowData.id)`

The third parameter is made optional (`id?: string`) in the `ColumnConfigProps` interface so existing consumers using only two parameters remain unaffected.

### Screenshots / Videos

- Before 

https://github.com/user-attachments/assets/acaa67a7-d5e9-4e23-aca6-fdc48f998849

- After

https://github.com/user-attachments/assets/235488ce-6888-4ac5-a8b3-40abb3cebfc6

Signed-off-by: Rishi Raj <rishiraj@users.noreply.github.com>


src/custom/CatalogDesignTable/columnConfig.tsx
```diff
@@ -42,7 +42,7 @@ export const colViews: ColView[] = [
 interface ColumnConfigProps {
   handleShowDetails: (design: Pattern) => void;
   handleClone: (designId: string, name: string) => void;
-  handleCopyUrl: (designId: string, name: string) => void;
+  handleCopyUrl: (designId: string, name: string, id?: string) => void;
```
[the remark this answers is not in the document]
saurabhraghuvanshii: @rishiraj38 can you check this?

saurabhraghuvanshii (changes requested):

src/custom/CatalogDesignTable/columnConfig.tsx
```diff
@@ -263,7 +263,7 @@ export const createDesignColumns = ({
             },
             {
               title: 'Copy Link',
-              onClick: () => handleCopyUrl(rowData.id, rowData.name),
+              onClick: () => handleCopyUrl(rowData.id, rowData.name, rowData.id),
```
saurabhraghuvanshii, line 266: @rishiraj38 why two time row.Data.id ?

saurabhraghuvanshii (dismissed):

src/custom/CatalogDesignTable/columnConfig.tsx
rishiraj38: The MeshMap extension is still using an older signature that expects the ID as the 3rd argument (type, name, id). Because Sistent was only passing (id, name), that 3rd argument was ending up as undefined, which caused the bug. Passing the ID in both the 1st and 3rd spots fixes the bug for the extension while keeping it backward compatible with anything else that relies on the 1st argument.

rishiraj38: The ideal way to fix this would be updating the MeshMap extension directly to just read the ID from the 1st argument. But since the extension lives in a separate, private repo, doing the fix here in Sistent makes sure it works right away without waiting on another PR. Plus, if the extension ever gets updated to use the 1st argument instead of the 3rd, this change handles both cases, so it won't break again.

rishiraj38: https://github.com/layer5io/sistent/blob/master/src/custom/CatalogDesignTable/DesignTableColumnConfig.tsx#L29

saurabhraghuvanshii: Thanks @rishiraj38  lgtm!!

saurabhraghuvanshii (approved):

Rajesh-Nagarajan-11 (approved):

Junnygram (approved):

PR #1547, opened 2026-05-13
[refactor]: Modernize FlipCard Architecture

### LINKED ISSUE #1546: Refactor FlipCard to use CSS 3D Transforms and Explicit Props
### Context
The current FlipCard implementation relies on a JavaScript setTimeout to swap children during the animation. This causes several issues:

- Flicker/Blanking: Content can disappear if the JS thread is busy.

- Rigid API: Using an array for children is prone to runtime errors.

- Theming/Sizing: Difficult to pass dynamic backgrounds or calculate container sizes because the back face is not rendered until the flip occurs.

### Proposed Changes

- Replace array-based children with explicit frontElement and backElement props.

- Remove setTimeout logic and implement standard CSS rotateY(180deg) with backface-visibility: hidden.

- Add flipAction prop to support both 'click' and 'hover' interactions.

- Improve TypeScript safety by removing manual child-index checking.

Katotodan:
**Notes for Reviewers**
### Description
This PR refactors the FlipCard component to align with modern CSS 3D animation standards and improves the Developer Experience (DX).

#### Changes

1. **Logic**: Switched from a JS-timed visibility toggle to a CSS 3D transform stack. The front and back faces are now rendered simultaneously but positioned back-to-back.
Previously the background, and other properties were related to the theme (ThemeProvider); theme can be set differently based on a project. This PR aims to give a developer much control over the background, padding, size of the component from the project this component is imported to. 

2. **API:**

- Deprecated array-based children.

- Introduced frontElement and backElement for better clarity.

- Added flipAction (hover/click) to increase component versatility.

3. **Styling**: Used shouldForwardProp in MUI's styled utility to prevent custom props from leaking to the DOM.

### Testing Instructions

- Import FlipCard and pass two distinct components to frontElement and backElement.

- Toggle the flipAction prop between hover and click (set to click by default).

- Verify that the animation remains smooth even when the CPU is under load (since it is now GPU-accelerated).
- Ensure that the container have a reasonable size cause the FlipCard will take the entire size of its container.

This PR fixes #1546


Katotodan:
## Before

https://github.com/user-attachments/assets/6bee461b-c5bf-4748-9ad0-6b23e36332f9

As you can see from the video above, I was not able to change the background of the card from the project  I was working on. Setting also the height was diffucult, same as padding.
This PR aims to solve that.

### After

https://github.com/user-attachments/assets/a96ac1c1-29d7-45be-a869-0433031011cd

arjunmehta-git:
Oops! Wrong button!

KatalKavya96 (commented):
@Katotodan Thanks for the refactor. The main direction looks good: `frontElement` / `backElement` makes the API safer, and removing the `setTimeout` based child swapping in favor of CSS 3D transforms is a good improvement.

One thing I think we should verify before merging: both `FrontContent` and `BackContent` are now absolutely positioned, while the outer `Card` depends on `height: 100%`.

Because of this, the card may not naturally derive its height from the rendered front/back content. This could still make dynamic sizing difficult which was one of the original problems mentioned in the issue.

Can we either:
1. ensure the component still sizes correctly based on its content or
2. clearly expose a way to pass explicit height/style/className to the container?

Apart from that the click/hover handling and removal of manual child indexing look aligned with the issue.

PR #1591, opened 2026-05-28
Standardize Icon Dimensions Using DEFAULT_WIDTH and DEFAULT_HEIGHT 

Rajesh-Nagarajan-11:
**Notes for Reviewers**

Ensured every icon consistently defaults to `DEFAULT_WIDTH` and `DEFAULT_HEIGHT`.

Many icons were either missing fallback values entirely or using hardcoded dimensions (such as `'24px'` or `'32px'`). This update standardizes them all for consistency and follows the same implementation pattern used across other icons.

### DIFF
--- a/src/icons/Copy/CopyIcon.tsx
+++ b/src/icons/Copy/CopyIcon.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { DEFAULT_WIDTH, DEFAULT_HEIGHT } from '../../constants/constants';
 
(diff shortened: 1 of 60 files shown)


src/icons/Copy/CopyIcon.tsx
rishiraj38, line 6: Shouldn't we make `width` and `height` optional here? Otherwise, since they're required, callers are still forced to pass them and they won't fallback to the new defaults.

PR #1592, opened 2026-05-28
Standardize Icon Dimensions Using DEFAULT_WIDTH and DEFAULT_HEIGHT

Rajesh-Nagarajan-11:
**Notes for Reviewers**

Ensured every icon consistently defaults to DEFAULT_WIDTH and DEFAULT_HEIGHT.

Many icons were either missing fallback values entirely or using hardcoded dimensions (such as '24px' or '32px'). This update standardizes them all for consistency and follows the same implementation pattern used across other icons.

### DIFF
--- a/src/icons/Calender/CalenderIcon.tsx
+++ b/src/icons/Calender/CalenderIcon.tsx
@@ -1,8 +1,9 @@
 import { FC } from 'react';
+import { DEFAULT_WIDTH, DEFAULT_HEIGHT } from '../../constants/constants';
 import { IconProps } from '../types';
 import { useTheme } from './../../theme/index';
 
-const CalenderIcon: FC<IconProps> = ({ width, height, ...props }) => {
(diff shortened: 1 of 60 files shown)


Sbragul26 (approved):
LGTM :+1:

Junnygram:
@Rajesh-Nagarajan-11  Replacing unique default dimensions with global defaults (`DEFAULT_WIDTH` / `DEFAULT_HEIGHT`) will cause layout shifts in consuming apps:
1. **Class Icons (`CommunityClassIcon`, `OfficialClassIcon`, `VerificationClassIcon`)**: These were custom-sized at `16x13` for compact inline display. Standardizing them to the global default (usually `24x24`) will make them much larger and potentially break inline alignments.
2. **Learning / Challenges Icons**: Downsized from their original `32px` default.
3. **GridView / TableView Icons**: Previously had a tall custom aspect ratio (`24x28.8`), now forced to standard square defaults.

Should we keep custom default constants for these specific sub-families of icons, or are we expecting consuming products to explicitly pass overrides at every usage site?

PR #1618, opened 2026-06-07
fix(modal): prevent fullscreen prop override

### LINKED ISSUE #1617: [Bug] Fullscreen toggle state overridden by external props in Modal component
### Current Behavior

The fullscreen toggle in the Modal component does not work correctly for consumers such as Meshery's Registry modal.

When the fullscreen icon is clicked:

- the toggle handler executes,
- the icon/tooltip changes between "Enter Fullscreen" and "Exit Fullscreen",
- internal fullscreen state updates correctly,

but the dialog itself never enters fullscreen mode.

Runtime DOM inspection shows that the MUI dialog never receives the MuiDialog-paperFullScreen class.

The issue appears to be caused by external props overriding the component's internal fullscreen state because ...props is spread after:

fullScreen={fullScreen}
fullWidth={!fullScreen}

### Expected Behavior

The fullscreen toggle should correctly switch the dialog into fullscreen mode.

After clicking the fullscreen icon:

- the dialog should visually expand to fullscreen,
- the DOM should contain the MuiDialog-paperFullScreen class,
- internal fullscreen state should not be overridden by external props.

### Screenshots/Logs

- Runtime verification:

document.querySelector('.MuiDialog-paper').className

- Observed result after clicking fullscreen:

MuiPaper-root MuiDialog-paper MuiDialog-paperWidthMd MuiDialog-paperFullWidth

- Expected result:

MuiDialog-paperFullScreen

### Environment

Browser: Google Chrome
OS: Ubuntu (WSL2)\
Framework: React + MUI
Repository: Sistent
Issue observed in: Meshery Registry modal

---

- 🎨 Wireframes and [designs for Sistent site](https://www.figma.com/file/ekm0Hh6M25bSKpP5BcIJom/Sistent-Design-System?type=design&node-id=596%3A5227&mode=design&t=8MExPykHN1vQFCTZ-1) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)

### LINKED ISSUE #19830: [UI] Full Screen toggle does not work in Registry modal
## Description

The Full Screen toggle in the **Registry** modal is not functioning correctly. Clicking the Full Screen icon does not expand the modal to full-screen mode. Since the modal never enters full-screen mode, users are also unable to exit full-screen mode.

## Expected Behavior

Clicking the Full Screen icon should expand the Registry modal to occupy the available screen space. After entering full-screen mode, users should be able to click the same control to return to the default modal size.

## Actual Behavior

Clicking the Full Screen icon has no visible effect. The modal remains in its original size and does not enter full-screen mode.

## Screenshots / Recording

https://github.com/user-attachments/assets/aec00745-9ece-4f23-8cd7-cebd153f39bc

## Additional Context

The Full Screen control appears in the Registry modal UI but does not trigger the expected state change. As a result, users cannot access either full-screen mode or the exit full-screen functionality.

### Environment

- **Host OS:** Mac Linux Windows
- **Platform:** Docker or Kubernetes
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

Salmaan-M:
**Notes for Reviewers**
### Description

This PR fixes an issue where the Modal component's internal fullscreen state could be overridden by external props.

### Previously:

allowed consumers passing fullScreen or fullWidth props to override the internal fullscreen toggle state managed by the component itself.

This caused fullscreen toggle behavior to fail in downstream consumers such as Meshery's Registry modal.

### Changes Made
Filtered conflicting fullScreen and fullWidth props from external props
Preserved external overrides for unrelated props
Kept internal fullscreen state authoritative

### Testing
Verified fullscreen toggle works correctly locally
Verified dialog now receives MuiDialog-paperFullScreen
Confirmed no unintended changes to unrelated props

[Signed commits](https://github.com/layer5io/sistent/compare/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)

This PR fixes #1617 

- [X] Yes, I signed my commits.

### DIFF
--- a/src/custom/Modal/index.tsx
+++ b/src/custom/Modal/index.tsx
@@ -159,8 +159,10 @@ export const Modal: React.FC<ModalProps> = ({
   const toggleFullScreen = () => {
     setFullScreen((prev) => !prev);
   };
+const { fullScreen: _ignoredFullScreen, fullWidth: _ignoredFullWidth, ...restProps } = props;
   return (
     <StyledDialog
+      {...restProps}
       maxWidth={maxWidth}
       open={open}
       onClose={closeModal}
@@ -168,7 +170,6 @@ export const Modal: React.FC<ModalProps> = ({
       aria-describedby="alert-dialog-slide-description"
       fullScreen={fullScreen}
       fullWidth={!fullScreen}
-      {...props}
     >
       {title && (
         <ModalStyledHeader className="modal-header" data-testid="modal-header">
--- a/src/custom/Modal/index.tsx
+++ b/src/custom/Modal/index.tsx
@@ -14,6 +14,7 @@ interface ModalProps extends DialogProps {
   headerIcon?: React.ReactNode;
   reactNode?: React.ReactNode;
   isFullScreenModeAllowed?: boolean;
+  openFullscreenOnStart?: boolean;
 }
 
 interface ModalFooterProps {
@@ -153,13 +154,19 @@ export const Modal: React.FC<ModalProps> = ({
   children,
   maxWidth = 'xs',
   isFullScreenModeAllowed,
+  openFullscreenOnStart = false,
   ...props
 }) => {
-  const [fullScreen, setFullScreen] = useState(false);
+  const [fullScreen, setFullScreen] = useState(openFullscreenOnStart);
   const toggleFullScreen = () => {
     setFullScreen((prev) => !prev);
   };
-const { fullScreen: _ignoredFullScreen, fullWidth: _ignoredFullWidth, ...restProps } = props;
+  
+  const {
+  fullScreen: _ignoredFullScreen,
+  fullWidth: _ignoredFullWidth,
+  ...restProps
+  } = props;
   return (
     <StyledDialog
       {...restProps}
--- a/src/custom/Modal/index.tsx
+++ b/src/custom/Modal/index.tsx
@@ -14,7 +14,6 @@ interface ModalProps extends DialogProps {
   headerIcon?: React.ReactNode;
   reactNode?: React.ReactNode;
   isFullScreenModeAllowed?: boolean;
-  openFullscreenOnStart?: boolean;
 }
 
 interface ModalFooterProps {
@@ -154,19 +153,25 @@ export const Modal: React.FC<ModalProps> = ({
   children,
   maxWidth = 'xs',
   isFullScreenModeAllowed,
-  openFullscreenOnStart = false,
   ...props
 }) => {
-  const [fullScreen, setFullScreen] = useState(openFullscreenOnStart);
+  /*
+   * Use the incoming fullScreen prop only as the initial state.
+   * After initialization, fullscreen is managed internally so, external
+   * fullScreen prop values do not override user-triggered fullscreen toggles.
+   */
+  const {
+    fullScreen: initialFullScreenState = false,
+    fullWidth: _ignoredFullWidth,
+    ...restProps
+  } = props;
+
+  const [fullScreen, setFullScreen] = useState(initialFullScreenState);
+
   const toggleFullScreen = () => {
     setFullScreen((prev) => !prev);
   };
-  
-  const {
-  fullScreen: _ignoredFullScreen,
-  fullWidth: _ignoredFullWidth,
-  ...restProps
-  } = props;
+
   return (
     <StyledDialog
       {...restProps}


src/custom/Modal/index.tsx
```diff
@@ -159,16 +159,21 @@ export const Modal: React.FC<ModalProps> = ({
   const toggleFullScreen = () => {
     setFullScreen((prev) => !prev);
   };
+  const {
+  fullScreen: _ignoredFullScreen,
+  fullWidth: _ignoredFullWidth,
+  ...restProps
+} = props;
```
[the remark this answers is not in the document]
Salmaan-M: Improved readability and code style

saurabhraghuvanshii:
@Salmaan-M, any type of Loom recording demonstrating your changes would help us understand what you fixed and what the current behavior is.

Salmaan-M:
https://github.com/user-attachments/assets/c4f2a99a-ba9a-418c-9441-f87f4535f14b

This video demonstrates how prop reordering helps in fixing the [[UI] Full screen issue](https://github.com/meshery/meshery/issues/19830) @saurabhraghuvanshii

saurabhraghuvanshii (approved):

saurabhraghuvanshii:
Thanks @Salmaan-M lgtm!!

saurabhraghuvanshii:
learned new thing today from you.

Salmaan-M:
> learned new thing today from you.

what is it? 🙂 @saurabhraghuvanshii

saurabhraghuvanshii:
> > learned new thing today from you.
> 
> what is it? 🙂 @saurabhraghuvanshii

secret!!

Salmaan-M:
> > > learned new thing today from you.
> > 
> > 
> > what is it? 🙂 @saurabhraghuvanshii
> 
> secret!!

Come on yaarr? make me happy

banana-three-join (commented):
What happens if you don't ignore the fullwidth? Can you still open and close the modal? Because I believe the prop that is overriding the internal fullscreen serves as the initial state for whether the fullscreen should be open or closed on the first interation and then, the fullscreen state is handled by the component itself. This means that by ignoring the initial fullscreen prop, we might have a modal to start on fullscreen by default but since we are ignoring it, we can never have that initial state. (Which if that's the case, it's not stated in the prop type for the modal anyways so that would also need to be fixed)

Salmaan-M:
That's a very good point.

During debugging, what I observed was that the issue occurred because external `fullScreen={false}` continued overriding the internal fullscreen toggle state even after interaction.

My intention with filtering `fullScreen/fullWidth` was to prevent conflicting state ownership once the component starts managing fullscreen internally via:

```tsx
const [fullScreen, setFullScreen] = useState(false);
```

However, you're right that this also removes the possibility of consumers intentionally providing an initial fullscreen state through props.

I agree that there are really two separate concerns here:

1. Initial fullscreen state
2. Ongoing fullscreen control after internal toggle state takes over

If supporting an initial fullscreen state is desirable behavior, then a better approach may be:

* initialize internal state from the incoming prop once,
* but avoid continuously allowing external props to override the internal toggle state after interaction.

Something like:

```tsx
const [fullScreen, setFullScreen] = useState(props.fullScreen ?? false);
```

while still preventing later prop precedence conflicts.

I think this approach better preserves the initial-state capability while still addressing the runtime override issue that caused the fullscreen toggle bug. I'll prototype this approach and validate the behavior locally. What's your say @banana-three-join

banana-three-join:
@Salmaan-M yeah the idea seems about right but maybe the variable should have a more descriptive name like openFullscreenOnStart={boolean} instead of fullscreen

In regards of the desired behavior, that would be my best guest since as I mentioned earlier, the typing of the modal itself lacks a fullscreen prop. I would say that the prop also needs to be added to the typing so it's usage is a little bit clearer since at the end of the day, it has it's own internal state to manage whether the modal is in fullscreen mode or not

rishiraj38:
@Salmaan-M  FYI
https://discuss.layer5.io/t/how-to-test-sistent-components-for-development-environment/7532

Salmaan-M:
@banana-three-join That makes sense to me.

Using a more explicit prop like `openFullscreenOnStart` would separate the "initial fullscreen state" concern from the internal fullscreen toggle state much more clearly and avoid overloading the inherited MUI `fullScreen` prop semantics.

I also agree that the typing/API should make this behavior explicit instead of implicitly inheriting it through `DialogProps`.

I'll prototype this approach locally and validate the behavior in Meshery as well.

rishiraj38:
@Salmaan-M, how's the progress on the issue?

Salmaan-M:
I prototyped an ``openFullscreenOnStart`` prop locally. My thinking is that this makes the initialization behavior explicit while keeping fullscreen ownership inside the component. Does this align with the intended API direction before I update the implementation further? @banana-three-join @rishiraj38

banana-three-join:
Yep, go ahead and make the change @Salmaan-M

KhushamBansal:
@Salmaan-M Are you working on fixing this issue?

Salmaan-M:
@KhushamBansal  yes, I am working on it.

KhushamBansal:
@Salmaan-M Any updates??

Salmaan-M:
> @Salmaan-M Any updates??

I've completed the implementation changes in Sistent based on the latest review feedback. The fullscreen state is now initialized through the explicit openFullscreenOnStart prop rather than continuously relying on the inherited fullScreen prop, which avoids the internal/external state override that was causing the fullscreen toggle to stop working.

I'm currently validating the change in Meshery. While testing, I ran into a local development environment issue related to the linked Sistent package and Turbopack/module resolution, so I'm resolving that before I can complete end-to-end verification in the Registry modal.

I'll post another update once I've finished validation and can confirm the behavior in Meshery. @KhushamBansal

Salmaan-M:
Implemented the `openFullscreenOnStart` approach discussed earlier.

The modal now initializes its internal fullscreen state from an explicit `openFullscreenOnStart` prop while keeping fullscreen ownership within the component. This preserves the ability to start a modal in fullscreen mode while preventing external `fullScreen` / `fullWidth` props from continuously overriding the internal toggle state after interaction.

I've updated the implementation and pushed the changes for review. @banana-three-join

src/__testing__/hideRootObjectTitle.test.tsx
banana-three-join, line 1: Revert. This is of out of the scope for the fix.

.github/workflows/notify-dependents.yml
banana-three-join, line 1: Revert this is out of scope for the fix.

banana-three-join (commented):
Why the introduction of openFullscreenOnStart, now fullscreen shouldn't be utilized anymore so ``` const {
  fullScreen: _ignoredFullScreen,
  fullWidth: _ignoredFullWidth,
  ...restProps
  } = props; ``` serves as a guardrail for users still utilizing fullscreen. 

The problem is that there's a lot of components which are currently utilizing this configuration making it a breaking change due to builds not being able to proceed due to TS errors. 

Honestly the more I think about it, we might just be better off by just do the de-constructing, renaming the fullscreen prop to initialFullscreen pass it as a default value to the fullScreen state, and just pass the rest of the props to the StyledDialog because while small in paper, this change would require an update in every single component importing this component.

Or maybe someone has another idea so we don't go with a ducktape type fix.

banana-three-join (changes requested):

Salmaan-M:
That's a fair point. My reasoning for introducing `openFullscreenOnStart` was to separate the "initial fullscreen state" from the internally managed fullscreen state and make that behavior explicit.

However, I agree that introducing a new prop increases migration cost and would require updating existing consumers that currently rely on `fullScreen`.

Using the existing `fullScreen` prop only as the initial value for the internal state, while still de-structuring it before passing props to `StyledDialog`, seems like a cleaner non-breaking approach. That would preserve the current API while preventing external prop updates from continuously overriding the internal toggle state.

I'm happy to update the implementation in that direction if that's the preferred API. @banana-three-join

src/__testing__/hideRootObjectTitle.test.tsx
Salmaan-M: Removed the unrelated changes from this PR and kept the scope focused on the modal fix.

.github/workflows/notify-dependents.yml
Salmaan-M: Removed the unrelated changes from this PR and kept the scope focused on the modal fix.

Bhumikagarggg (approved):

banana-three-join (changes requested):
Do the following. 

Don't modify the interface for the time being or add the openFullscreenOnStart as a prop because the Modal component is being used throughout a lot of different areas in our different repos, so a change of variable would require for those repos to also be updated. 

Where you have ```const {
  fullScreen: _ignoredFullScreen,
  fullWidth: _ignoredFullWidth,
  ...restProps
  } = props;```

Instead, put it further up before ```const [fullScreen, setFullScreen] = useState(openFullscreenOnStart);``` 

Replace ```fullScreen: _ignoredFullScreen``` with ```fullScreen: initialFullScreenState``` and now pass it down as the initial value to ```const [fullScreen, setFullScreen] = useState(initialFullScreenState);``` 

Also, please add a /* comment */ on top of this behavior that briefly explains why this approach was added

Salmaan-M:
@banana-three-join Updated the implementation to keep the existing `fullScreen` API and use it only as the initial fullscreen state. Added an inline comment explaining the behavior and removed the `openFullscreenOnStart` prop to avoid introducing a breaking change.

banana-three-join (approved):

banana-three-join:
@Salmaan-M Ty!

rishiraj38 (approved):

saurabhraghuvanshii (approved):

saurabhraghuvanshii:
Thanks @Salmaan-M  lgtm!!

PR #1622, opened 2026-06-11
fix(theme): use default checkbox icon components

### LINKED ISSUE #1621: [Bug] Checkbox root transparency makes custom icons invisible
**Description**
In `@sistent/sistent's` theme overrides (`src/theme/components/checkbox.modifier.ts`), color: `transparent` is applied globally at the root level of `MuiCheckbox.`

Because standard SVG icons inherit color via `currentColor`, any custom SVG icon component passed to the `icon` or `checkedIcon` props of a Checkbox (e.g. `FilterAllIcon` used in Meshery's cluster dropdown) automatically inherits this root transparency and becomes completely invisible in its unchecked state.

**Steps to Reproduce**
1. Import and render a Checkbox from Sistent/MUI.
2. Pass a custom SVG icon (without hardcoding an inline fill attribute) to the icon prop.
3. Observe that the custom icon path is completely transparent/invisible.

**Expected Behaviour** 
1. Default Sistent checkboxes should keep their custom border and hide the default outline path.
2. Custom icons passed as props should be fully visible and respect default state/theme colors instead of inheriting transparent.

---

- 🎨 Wireframes and [designs for Sistent site](https://www.figma.com/file/ekm0Hh6M25bSKpP5BcIJom/Sistent-Design-System?type=design&node-id=596%3A5227&mode=design&t=8MExPykHN1vQFCTZ-1) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)

kanishksingh23:
**Notes for Reviewers**
This PR resolves the checkbox color bleeding bug by adding custom default SVG icons. It ensures that custom child icons (like `FilterAllIcon` in Meshery UI) are fully visible and inherit the correct colors, while standard checkboxes preserve Sistent's exact border design.

This PR fixes #1621 

### DIFF
--- a/src/icons/Checkbox/CheckboxCheckedIcon.tsx
+++ b/src/icons/Checkbox/CheckboxCheckedIcon.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { DEFAULT_HEIGHT, DEFAULT_WIDTH } from '../../constants/constants';
+import { IconProps } from '../types';
+
+export const CheckboxCheckedIcon = ({
+  width = DEFAULT_WIDTH,
+  height = DEFAULT_HEIGHT,
+  ...props
+}: IconProps): JSX.Element => (
+  <svg
+    width={width}
+    height={height}
+    viewBox="0 0 24 24"
+    xmlns="http://www.w3.org/2000/svg"
+    {...props}
+    fill="currentColor"
+  >
+    <path d="M19 3H5c-1.11 0-2 .9-2 2v14c0 1.1.89 2 2 2h14c1.11 0 2-.9 2-2V5c0-1.1-.89-2-2-2zm-9 14l-5-5 1.41-1.41L10 14.17l7.59-7.59L19 8l-9 9z" />
+  </svg>
+);
+
+export default CheckboxCheckedIcon;
--- a/src/icons/Checkbox/CheckboxCheckedIcon.tsx
+++ b/src/icons/Checkbox/CheckboxCheckedIcon.tsx
@@ -5,6 +5,7 @@ import { IconProps } from '../types';
 export const CheckboxCheckedIcon = ({
   width = DEFAULT_WIDTH,
   height = DEFAULT_HEIGHT,
+  fill = 'currentColor',
   ...props
 }: IconProps): JSX.Element => (
   <svg
@@ -12,8 +13,8 @@ export const CheckboxCheckedIcon = ({
     height={height}
     viewBox="0 0 24 24"
     xmlns="http://www.w3.org/2000/svg"
+    fill={fill}
     {...props}
-    fill="currentColor"
   >
     <path d="M19 3H5c-1.11 0-2 .9-2 2v14c0 1.1.89 2 2 2h14c1.11 0 2-.9 2-2V5c0-1.1-.89-2-2-2zm-9 14l-5-5 1.41-1.41L10 14.17l7.59-7.59L19 8l-9 9z" />
   </svg>
--- a/src/icons/Checkbox/CheckboxIcon.tsx
+++ b/src/icons/Checkbox/CheckboxIcon.tsx
@@ -5,6 +5,7 @@ import { IconProps } from '../types';
 export const CheckboxIcon = ({
   width = DEFAULT_WIDTH,
   height = DEFAULT_HEIGHT,
+  stroke = 'currentColor',
   ...props
 }: IconProps): JSX.Element => (
   <svg
@@ -12,8 +13,8 @@ export const CheckboxIcon = ({
     height={height}
     viewBox="0 0 24 24"
     xmlns="http://www.w3.org/2000/svg"
-    {...props}
     fill="none"
+    {...props}
   >
     <rect
       x="3"
@@ -21,7 +22,7 @@ export const CheckboxIcon = ({
       width="18"
       height="18"
       rx="2"
-      stroke="currentColor"
+      stroke={stroke}
       strokeWidth="1.5"
     />
   </svg>
--- a/src/icons/Checkbox/CheckboxIcon.tsx
+++ b/src/icons/Checkbox/CheckboxIcon.tsx
@@ -2,12 +2,16 @@ import React from 'react';
 import { DEFAULT_HEIGHT, DEFAULT_WIDTH } from '../../constants/constants';
 import { IconProps } from '../types';
 
+type CheckboxIconProps = IconProps & {
+  innerRectStroke?: string;
+};
+
 export const CheckboxIcon = ({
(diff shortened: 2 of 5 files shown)


src/theme/components/checkbox.modifier.ts
```diff
@@ -27,6 +26,9 @@ export const MuiCheckbox: Components<Theme>['MuiCheckbox'] = {
             padding: '0px'
           }
         },
+        '&:not(.Mui-checked):not(.Mui-indeterminate) .MuiSvgIcon-root': {
+          color: 'transparent'
+        },
```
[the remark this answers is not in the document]
kanishksingh23: Thanks for the suggestion! That is a very clean architectural approach.

Currently, Sistent does not have custom unchecked/checked SVG components in the repository, and using defaultProps would require bundling custom SVG paths/components with the theme.

For now, this scoped CSS selector: &:not(.Mui-checked): not(.Mui-indeterminate). MuiSvgIcon-root serves as a pragmatic, CSS-only fix to isolate the transparency override as much as possible to the default unchecked state. Since custom icons used in our components (like FilterAllIcon) define explicit fills/styles, they successfully bypass this override.

banana-three-join (commented):
I feel like this solution look good in paper but it also brings another set of issues. What's going to happen to the SVGs that inherit this "color: transparent" property in "fill: currentColor"? If the parent doesn't has a color attribute defined, it'll go to the grandparent searching for it, causing a series of unpredictable issues. Also, it's not easy to debug when the items are using this pseudo class so that's another issue. It might be a better approach to look into creating a DefaultUncheckedIcon but the issue is that MUI handles it's own icons when don't being checked so it'll be good to see how MUI handles their to implement the same solution so all of the icons have the same behavior to them once unchecked even if they're custom ones.

kanishksingh23:
@banana-three-join, I agree that using pseudo-classes (: not(.Mui-checked)) is hard to debug and can cause inheritance issues, where child elements look up the grandparent chain for their colour.

As you suggested, I am going to implement the DefaultUncheckedIcon approach instead, completely removing the CSS transparency and border tricks:
1. Create Default Icons: Create CheckboxIcon and CheckboxCheckedIcon inside Sistent's assets.
2. Set via defaultProps: Register these icons under MuiCheckbox.defaultProps in the theme modifier.

This ensures that:
1. Standard checkboxes render Sistent's exact border design using the native SVG path.
2. Custom icons passed down by consumers are fully visible and inherit color/fill properly.

leecalcote:
The checkbox on the "Where would you like to start" modal in Layer5 Cloud might be an example of where checkbox mis-coloring occurs.

kanishksingh23:
@leecalcote, thank you for pointing this out. Yes, since the color: 'transparent' override was applied globally to the checkbox root in Sistent, any checkbox in Layer5 Cloud that relied on custom icons or standard outlines would exhibit this mis-coloring/invisibility bug.

This PR removes that global transparency override and replaces it with proper default SVG icons, so it will resolve the invisibilty/mis-coloring bug.

kanishksingh23:
I have verified the changes locally.

Unchecked state:
Screenshot 2026-06-12 at 11 47 18 PM

Checked state:
Screenshot 2026-06-12 at 11 47 34 PM

src/icons/Checkbox/CheckboxCheckedIcon.tsx
```diff
@@ -0,0 +1,22 @@
...
+  height = DEFAULT_HEIGHT,
+  ...props
+}: IconProps): JSX.Element => (
+  <svg
+    width={width}
+    height={height}
+    viewBox="0 0 24 24"
+    xmlns="http://www.w3.org/2000/svg"
+    {...props}
+    fill="currentColor"
```
banana-three-join: Just like width and height, fill should receive a fill prop with a default value of "currentColor"

src/icons/Checkbox/CheckboxIcon.tsx
```diff
@@ -0,0 +1,30 @@
...
+    {...props}
+    fill="none"
+  >
+    <rect
+      x="3"
+      y="3"
+      width="18"
+      height="18"
+      rx="2"
+      stroke="currentColor"
```
banana-three-join: Stroke shouldn't only depend on "currentColor" to create the outline for the CheckboxIcon. While a good default value, there might be a case in which the nearest color provider for rect to inherit might have an undesired color. 

Allow for a prop to be defined to determine the color of the stroke make currentColor it's default value.

banana-three-join (changes requested):

src/icons/Checkbox/CheckboxIcon.tsx
kanishksingh23: @banana-three-join, Done! Added the stroke prop to the component with a default value of 'currentColor' and passed it to the <rect> tag.

src/icons/Checkbox/CheckboxCheckedIcon.tsx
kanishksingh23: @banana-three-join, Done! Added the fill prop to the component with a default value of 'currentColor' and passed it to the <svg> tag.

src/icons/Checkbox/CheckboxIcon.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { DEFAULT_HEIGHT, DEFAULT_WIDTH } from '../../constants/constants';
+import { IconProps } from '../types';
+
+export const CheckboxIcon = ({
+  width = DEFAULT_WIDTH,
+  height = DEFAULT_HEIGHT,
+  stroke = 'currentColor',
```
banana-three-join: Could you rename the variable to something which reflects that the attribute being modified is the stroke of the rect inside and the svg itself? Something kinda like innerRectStroke or something along those lines.

banana-three-join (changes requested):
Aside from the small change, I would say it looks good!

src/icons/Checkbox/CheckboxIcon.tsx
kanishksingh23: @banana-three-join, Done!

kanishksingh23:
@banana-three-join, I have also updated the PR title and description to match the latest changes.

banana-three-join (approved):

banana-three-join:
I think it looks good so far. Could you send over a video to take a look at its behavior?

kanishksingh23:
@banana-three-join, here is the video to showcase the changes.

https://github.com/user-attachments/assets/44b9756f-9f44-48f5-a8d5-e1bfc0d9565b

PR #1633, opened 2026-06-15
fix: align InviteUserModal payload structure with schema definition

rishiraj38:
## Description

This PR fixes part of https://github.com/layer5io/meshery-cloud/issues/5484.

While testing the invite user flow, I found that the [generated RTK hook expects](https://github.com/meshery/schemas/blob/master/typescript/rtk/cloud.ts#L8839) the payload in the following format:

```ts
{
  orgId: string;
  body: { ... }
}
```

However, the UI was passing:

```ts
{
  orgId: string;
  userInvite: { ... }
}
```

Because of this mismatch, the generated hook was unable to populate the request body correctly, resulting in a "cannot parse request body" error.

Since the schema is the source of truth, this PR keeps the schema unchanged and adds a wrapper in Meshery Cloud to adapt the payload to the format expected by the generated hook.

After making this change and testing with the updated schemas, the request body is now sent correctly and the previous "cannot parse request body" error no longer occurs.

During verification, the error changed to "failed persisting user record", confirming that the request is now reaching the backend successfully. The issue causing that error is addressed separately in a follow-up PR.

## Changes

* Updated the user invite mutation payload to use `body` instead of `userInvite`, aligning it with the payload structure expected by the schema-generated RTK hook.
* Fixed the payload mismatch that prevented the request body from being sent correctly to the backend.
* Verified that the request now reaches the backend successfully and no longer results in a request body parsing error.

## Before 

https://github.com/user-attachments/assets/4adf46ba-9636-495e-8c03-a17e4bac0ec3

## After

https://github.com/user-attachments/assets/2ec853ba-4a5a-445a-bbec-5c61b6ed9116

### DIFF
--- a/src/custom/DashboardWidgets/GettingStartedWidget/InviteUserModal.tsx
+++ b/src/custom/DashboardWidgets/GettingStartedWidget/InviteUserModal.tsx
@@ -210,7 +210,7 @@ export default function UserInviteModal({
 
     try {
       await userInvite({
-        userInvite: {
+        body: {
           first_name: inviteeFirstName,
           last_name: inviteeLastName,
           email: inviteeEmail,


banana-three-join (approved):
LGTM

PR #1691, opened 2026-07-07
fix: reduce CustomTooltip delays to eliminate sluggishness

shashank-tomar0:
**Description**
Reduces the hardcoded enter/leave delays in `CustomTooltip` that cause sluggish tooltip behavior.

**Changes**
- `enterDelay`: `150` → `100` (configurable via props, default `100`)
- `leaveDelay`: `700` → `0` (configurable via props, default `0`) — this was the main cause of sluggishness
- Removed `enterNextDelay={400}` (unnecessary with `leaveDelay=0`)
- All delay values are now overridable by callers via props

**Related**
Companion PR in meshery/meshery: adds theme-level `MuiTooltip.defaultProps` for direct `Tooltip` usages not using `CustomTooltip`.

### DIFF
--- a/src/custom/CustomTooltip/customTooltip.tsx
+++ b/src/custom/CustomTooltip/customTooltip.tsx
@@ -27,15 +27,16 @@ function CustomTooltip({
   variant = 'standard',
   bgColor = '#141414',
   slotProps = {},
+  enterDelay = 100,
+  leaveDelay = 0,
   ...props
 }: CustomTooltipProps): JSX.Element {
   const theme = useTheme();
 
   return (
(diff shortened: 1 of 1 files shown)


leecalcote (changes requested):
Thanks, @shashank-tomar0.


1. sign off on your commits, please.
2. There's a tradeoff here. Too fast to display the tooltip and that, too, is annoying. Capture short videos of before and after your changes. Include in the PR description.

rishiraj38:
Fixed https://github.com/layer5io/sistent/pull/1699.

PR #1704, opened 2026-07-10
feat: integrate PermissionProvider and permission hooks

rishiraj38:
#### **Summary**
This PR implements a generic, CASL-agnostic `PermissionProvider` in Sistent and integrates it into Meshery UI. By moving permission checks into Sistent's base components and checking them declaratively via props, we decouple Sistent from direct CASL dependency and eliminate hardcoded `sessionStorage` lookups from within Sistent.

---

##### **Sistent Library (`/sistent`)**
1. **`PermissionProvider` (`src/custom/PermissionProvider.tsx`)**
   - Added a new CASL-agnostic context/hook wrapper. It accepts a generic `userHasPermission: (key: Key) => boolean` callback and optional `userContext` information.
2. **`PermissionShield` (`src/custom/permissions.tsx`)**
   - Removed direct dependency on `sessionStorage` (e.g., `sessionStorage.getItem('currentOrg')`). 
   - Now reads `userName`, `orgName`, and `roleNames` directly from `PermissionProvider`'s `userContext`.
3. **Base Components (`Button`, `IconButton`, `MenuItem`, `ListItem`, `ListItemButton`)**
   - Updated components to accept `permissionKey` (from `@meshery/schemas`) and `permissionAction: 'showShield' | 'hide'`.
   - The components automatically evaluate permissions via the provider hook and apply the appropriate action if the user lacks the permission (e.g., automatically rendering the `PermissionShield` tooltip overlay).

##### **[Meshery UI (`/meshery/ui`)](https://github.com/meshery/meshery/pull/20590)**
1. **`_app.tsx`**
   - Mounted the `PermissionProvider` at the root, passing down a CASL adapter (`ability.can`) and feeding `userContext` from the Redux store/queries.
2. **`Navigator.tsx` & `navigatorComponents.tsx`**
   - Converted Navigator item configurations to pass the canonical `Keys` directly.
   - Removed custom `CAN()` imports and calls inside `Navigator.tsx` in favor of declarative `permissionKey` and `permissionAction` props on `ListItemComponent`.

---

#### **Benefits**
- **Decoupled Architecture**: Sistent and its base components are now completely unaware of CASL. Transitioning away from CASL in the future will require changing only one function inside `_app.tsx`.
- **Cleaner Code**: Eliminates redundant `disabled={!CAN(...)}` logic and wrapper elements on the caller's side.
- **No Storage Leakage**: Removed direct dependencies on `sessionStorage` inside the UI library.

### DIFF
--- a/src/base/Button/Button.tsx
+++ b/src/base/Button/Button.tsx
@@ -12,7 +12,6 @@ export interface ButtonProps extends MuiButtonProps {
    *
    * - `'showShield'` (default) — disables the button and shows a shield icon
    *   with a permission-metadata tooltip.
-   * - `'disable'` — disables the button without a shield.
    * - `'hide'` — renders nothing.
    *
    * Ignored when `permissionKey` is not provided.
@@ -30,18 +29,8 @@ export function Button({
 }: ButtonProps): JSX.Element {
   const hasPermission = useHasPermission(permissionKey);
 
-  // No permissionKey → normal behavior (backward compatible)
-  if (!permissionKey) {
-    return (
-      <MuiButton {...props} disabled={disabled}>
-        {label}
-        {children}
-      </MuiButton>
-    );
-  }
-
-  // User HAS permission → render normally
-  if (hasPermission) {
+  // No permissionKey or user has permission → render normally
+  if (!permissionKey || hasPermission) {
     return (
       <MuiButton {...props} disabled={disabled}>
         {label}
@@ -51,27 +40,18 @@ export function Button({
   }
 
   // User LACKS permission → apply the permissionAction
-  switch (permissionAction) {
-    case 'hide':
-      return <></>;
-    case 'disable':
-      return (
-        <MuiButton {...props} disabled={true}>
-          {label}
-          {children}
-        </MuiButton>
-      );
-    case 'showShield':
-    default:
-      return (
-        <PermissionShield permissionKey={permissionKey} variant="badge">
-          <MuiButton {...props} disabled={true}>
-            {label}
-            {children}
-          </MuiButton>
-        </PermissionShield>
-      );
+  if (permissionAction === 'hide') {
+    return <></>;
   }
+
+  return (
+    <PermissionShield permissionKey={permissionKey} variant="badge">
+      <MuiButton {...props} disabled={true}>
+        {label}
+        {children}
+      </MuiButton>
+    </PermissionShield>
+  );
 }
 
 export const ContainedButton = (props: ButtonProps): JSX.Element => (
--- a/src/base/Button/Button.tsx
+++ b/src/base/Button/Button.tsx
@@ -1,38 +1,77 @@
 import { Button as MuiButton, type ButtonProps as MuiButtonProps } from '@mui/material';
 import React from 'react';
+import { useHasPermission, type PermissionAction } from '../../custom/PermissionProvider';
 import { Key, PermissionShield } from '../../custom/permissions';
 
 export interface ButtonProps extends MuiButtonProps {
   label?: string;
   children?: React.ReactNode;
   permissionKey?: Key;
+  /**
+   * Determines behavior when the user lacks the required permission.
+   *
+   * - `'showShield'` (default) — disables the button and shows a shield icon
+   *   with a permission-metadata tooltip.
+   * - `'disable'` — disables the button without a shield.
+   * - `'hide'` — renders nothing.
+   *
+   * Ignored when `permissionKey` is not provided.
+   */
+  permissionAction?: PermissionAction;
 }
 
 export function Button({
   label,
   children,
   permissionKey,
+  permissionAction = 'showShield',
   disabled,
   ...props
 }: ButtonProps): JSX.Element {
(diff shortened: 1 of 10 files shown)


banana-three-join (changes requested):
```
  if (!permissionKey) {
    return <MuiListItem {...rest} ref={ref} />;
  }

  // User HAS permission → render normally
  if (hasPermission) {
    return <MuiListItem {...rest} ref={ref} />;
  }
```

These conditional evaluations are contradictory and should be avoided. A user without a permission key shouldn't have access to the same layouts and components as someone with a permission key. Instead, the components should be explicitly rendered in either a hidden, disabled or visible state.

On that note, I believe there isn't much that much of a difference between the disabled and show guard state so the disabled state should be removed.

src/custom/PermissionProvider.tsx
```diff
@@ -0,0 +1,144 @@
...
+  roleNames?: string[];
+}
+
+/**
+ * The `Key` interface mirrors the structure defined in `@meshery/schemas`.
+ *
+ * Re-declared here so that `PermissionProvider` can reference it without
+ * importing from the permissions barrel (which would create a circular dep).
+ */
+export interface PermissionKey {
```
mrmerlin320: this should be from schemas

rishiraj38: Please see the comment above. This is intentionally re-declared here to avoid a circular dependency. Importing it from `schemas` through the permissions barrel would introduce that cycle.

rishiraj38: We can import this from schemas, but simply removing `subject` and `action` might cause issues since they're still there for backwards compatibility with CanShow.

aabidsofi19: Subject and action should not part of permission key .  The permission handler in meshery ui which uses casl should use the key id  for it

rishiraj38: Updated: Key is now imported directly from schemas instead of being re-declared locally.

Also cleaned up `subject` and `action` from the `Key` interface, since they are only kept as optional fields in `HasKeyProps` for backward compatibility with existing `CanShow`usage; the core `Key` type is now purely from schemas.

rishiraj38:
> ```
>   if (!permissionKey) {
>     return <MuiListItem {...rest} ref={ref} />;
>   }
> 
>   // User HAS permission → render normally
>   if (hasPermission) {
>     return <MuiListItem {...rest} ref={ref} />;
>   }
> ```
> 
> These conditional evaluations are contradictory and should be avoided. A user without a permission key shouldn't have access to the same layouts and components as someone with a permission key. Instead, the components should be explicitly rendered in either a hidden, disabled or visible state.
> 
> On that note, I believe there isn't much that much of a difference between the disabled and show guard state so the disabled state should be removed.

Good catch! Collapsed the two early returns into a single `if (!permissionKey || hasPermission)` guard and removed the `disable` action entirely; showShield already communicates why the action is restricted, so disable was just a worse version of it. PermissionAction is now just 'hide' | 'showShield'.

src/base/Button/Button.tsx
```diff
@@ -1,37 +1,56 @@
...
   label,
   children,
   permissionKey,
+  permissionAction = 'showShield',
   disabled,
   ...props
 }: ButtonProps): JSX.Element {
-  // When disabled AND permissionKey is provided, show the shield overlay
-  if (disabled && permissionKey) {
+  const hasPermission = useHasPermission(permissionKey);
```
banana-three-join, line 30: I believe useHasPermission should be fully responsible of determining whether an user is able to see the Button or not. If there's no permissionKey given, the user shouldn't have the needed permission to click on the button. This would allow to remove the !permissionKey evaluation further up and streamline the condition to: 

if the user hasPermission -> render it normally, if not true, then choose from the permission action to either hide it or use the badge.

If a user doesn't has a permissionKey, they shouln't be able to see the same buttons as an user with a permissionKey, they should be treated just like a user without permissions.

Also, as it is right now, if a user has permission, they are still not going to be able to see the item, because of the permissionAction conditional restricts them from doing so. When, I believe the purpose of the permissionAction is to determine the behavior of how will the button be disabled, either hide it or show the badge on top of it.

And for last, even if a button can be rendered, I believe it should still be wrapped by the permissionBadge so the permissionKey metadata isn't lost on approval.

banana-three-join (dismissed):

src/base/Button/Button.tsx
rishiraj38: Makes sense! Removed the !permissionKey guard; useHasPermission now handles that internally (returns true when no key is given), so the flow is cleaner: hasPermission -- render normally, otherwise --  hide or shield. permissionAction never runs for permitted users.

On always wrapping in PermissionShield for metadata visibility, cool idea, but it'd need a "permitted" render mode in PermissionShield (right now it always dims + disables). Will track that as a follow-up.

rishiraj38:
@aabidsofi19  @banana-three-join, any more feedback on the PR, or are we good to merge?

banana-three-join (approved):

Issue #1731, opened 2026-07-21
Place the Permission Shield at the Bottom-Left of Disabled Components


rishiraj38:
### Current Behavior

The permission shield icon is currently positioned in the top-right corner of disabled components.

### Expected Behavior

Move the permission shield to the bottom-left corner of the disabled component.

### Screenshots/Logs

<img width="241" height="117" alt="Image" src="https://github.com/user-attachments/assets/23eed7b1-3796-495f-ae7f-8b67c11ee0fd" />

### Environment

---

#### Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources

- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/build-and-release)
- 📚 [Instructions for contributing to documentation](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#documentation-contribution-flow)
  - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Sistent site](https://www.figma.com/file/ekm0Hh6M25bSKpP5BcIJom/Sistent-Design-System?type=design&node-id=596%3A5227&mode=design&t=8MExPykHN1vQFCTZ-1) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

Salmaan-M:
@rishiraj38 can you assign me on this?

rishiraj38:
@Salmaan-M, any updates ?

Salmaan-M:
@rishiraj38 working on it. Sorry for the delay

PR #1732, opened 2026-07-22
Extensions: Compatibility guarantee

marblom007:
## What this is

The single declaration of the Meshery host <-> extension contract, which Meshery UI and Kanvas both derive from instead of hand-duplicating string literals.

This is the keystone of a three-repo change:

| Repo | PR | Role |
|---|---|---|
| `layer5io/sistent` | **this PR** | declares the contract |
| `meshery/meshery` | #20880 | host half - types its bus, injects `contractVersion` |
| `layer5labs/meshery-extensions` | #4303 | Kanvas half - asserts the injected bag at mount |

Both consumers need a published `@sistent/sistent` carrying `src/actors/mesheryExtensionContract.ts`, so this lands and releases first.

## Why

Three things cross the host/extension boundary: the bundle's export shape, a bag of injected capabilities handed over as `injectProps`, and an RxJS event bus. The last two were duplicated string literals in each repo, and the host bus was declared as a bare `new EventBus()` - whose type parameter collapses to its constraint, so `publish()` accepted any `{ type: string }`.

A rename on either side therefore compiled cleanly on **both** and failed silently at runtime: the subscriber's `if (event.type === ...)` simply stopped matching and the feature became a no-op with no error, no warning, and no failing test. That is how `OPEN_DESIGN_IN_KANVAS` -> `OPEN_DESIGN_IN_EXTENSION` shipped broken, and how `capabilitiesRegistry` -> `providerCapabilities` shipped before it.

Deriving every literal from one declaration turns that class of rename into a compile error in every consumer.

## What it declares

- `MesheryExtensionEvent` - the event union, and `MesheryExtensionEventBus` as the bus alias both sides must use
- `MESHERY_EXTENSION_EVENT` - named handles for every literal, with a compile-time guard so the map cannot drift behind the union
- `MESHERY_EXTENSION_CAPABILITIES` / `MESHERY_EXTENSION_HOOKS` - keys the host is expected to put on `injectProps`
- `MESHERY_EXTENSION_DEPRECATED_CAPABILITIES` - keys the host still injects so already-published bundles keep mounting
- `MESHERY_EXTENSION_CONTRACT_VERSION` - the handshake value

### Compile-time agreement is necessary but not sufficient

Extension bundles are published artifacts loaded into whichever host version is deployed, so host and extension are never guaranteed to be built from the same commit. No CI gate can observe that skew. The runtime helpers exist to make it loud:

- `isMesheryExtensionEvent` - narrows a bus event, checking both the discriminant **and** the payload it promises to carry, so a renamed payload field is caught rather than handed to a subscriber that reads `undefined` off it. Extra fields are tolerated, so additive changes stay compatible.
- `reportInjectedCapabilities` / `isInjectedCapabilityReportSatisfied` / `describeInjectedCapabilityReport` - compare a host's `injectProps` against the contract and render one actionable message naming the offending keys. Side-effect free, so the host asserts on it in a unit test (catching a prop rename before merge) and extensions run it at mount (catching deploy-time skew).

A `contractVersion` the host advertises that differs from this build's fails the check. An **absent** version is the supported legacy case - hosts deployed before the handshake existed predate the field rather than disagreeing about it, and flagging them all at once would train maintainers to ignore the report.

## Why the DateTimePicker change is in this PR

Not a drive-by. `@mui/x-date-pickers` is declared an **optional** peer dependency, but `DateTimePicker.tsx` imported it at module scope - and that module is reachable from the package barrel. So merely writing `import { anything } from '@sistent/sistent'` executed a top-level require of a package most consumers never install, throwing `Cannot find module '@mui/x-date-pickers/AdapterDateFns'` with a message pointing at sistent rather than at the missing peer. On a clean install it took out four unrelated Kanvas Jest suites.

That is precisely the import Kanvas has to make to consume the contract, so **this fix is what makes the contract consumable at all**. Resolving the picker on first render via `React.lazy` keeps the optional dependency genuinely optional: the cost is paid only by code that actually renders one, which is what `peerDependenciesMeta.optional` already advertises.

Verified at the bundle level - the built CJS/ESM output contains only dynamic `import()` of the three picker entrypoints and no top-level `require`.

## Also in here

`createCanShow` is now exported explicitly from the package entrypoint. It was reachable at runtime but dropped from the generated `.d.ts` by a nested-barrel quirk, so consumers silently got `any` - which meant its `eventBus` argument was not variance-checked at the one place a host hands its bus to sistent. Same quirk as `FeedbackButton` above it.

`MissingPermissionReason` / `MissingCapabilityReason` moved from `custom/permissions.tsx` into the contract and are re-exported, so **every existing import path keeps working**. They moved because `PermissionShield` publishes them onto whichever bus the caller supplies, which in Meshery *is* the host <-> extension bus - making them part of the contract whether or not we had declared them so.

## Compatibility

Additive. No existing export was removed or renamed, and no runtime behaviour changed outside `DateTimePicker`'s resolution timing. Per the file's own rule, additive changes need no `MESHERY_EXTENSION_CONTRACT_VERSION` bump, so it stays at `1`.

## Testing

23 Jest cases in `src/__testing__/mesheryExtensionContract.test.ts`, covering the wire literals (pinned by hand on purpose - a rename is only safe alongside a version bump and a deprecation entry, and this test is what forces that conversation), malformed and renamed payloads for known literals, a valid payload for every literal, extra-field tolerance, missing capabilities and hooks, deprecated aliases, unrecognized keys, and the version handshake across matching / mismatched / non-numeric / absent.

Full suite green (411 tests, 22 suites), `eslint .` clean, `npm run build` clean.

Signed-off-by: Ahmed Jamil <197998703+CodeAhmedJamil@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * Added a shared Meshery extension contract with version handshakes, event validation, capability reporting, and compatibility diagnostics.
  * Exposed extension contract types and utilities through the public API.
  * Improved generated type declarations for permission helpers.

* **Bug Fixes**
  * Date and time picker dependencies now load only when needed, preventing failures when optional packages are unavailable.

* **Tests**
  * Added comprehensive coverage for extension events, capabilities, deprecated interfaces, and contract-version compatibility.

### DIFF
--- a/src/__testing__/mesheryExtensionContract.test.ts
+++ b/src/__testing__/mesheryExtensionContract.test.ts
@@ -1,0 +1,175 @@
+import {
+  MESHERY_EXTENSION_CAPABILITIES,
+  MESHERY_EXTENSION_CONTRACT_VERSION,
+  MESHERY_EXTENSION_DEPRECATED_CAPABILITIES,
+  MESHERY_EXTENSION_EVENT,
+  MESHERY_EXTENSION_EVENT_TYPES,
+  MESHERY_EXTENSION_HOOKS,
+  describeInjectedCapabilityReport,
+  isInjectedCapabilityReportSatisfied,
+  isMesheryExtensionEvent,
+  reportInjectedCapabilities
+} from '../actors/mesheryExtensionContract';
+
+/** A host bag that satisfies the contract exactly, built from the contract itself. */
+const satisfyingInjectProps = () => ({
+  ...Object.fromEntries(MESHERY_EXTENSION_CAPABILITIES.map(key => [key, () => null])),
+  hooks: Object.fromEntries(MESHERY_EXTENSION_HOOKS.map(hook => [hook, () => null]))
+});
+
+describe('Meshery extension contract — event literals', () => {
+  it('exposes a literal for every event type on the bus', () => {
+    // The union -> map completeness guard is compile-time; this asserts the
+    // runtime array stays in step with the map it is derived from.
+    expect([...MESHERY_EXTENSION_EVENT_TYPES].sort()).toEqual(
+      Object.values(MESHERY_EXTENSION_EVENT).sort()
+    );
+  });
+
+  it('pins the wire literals that cross the host boundary', () => {
+    // Hard-coded on purpose. These strings are the actual contract: a rename is
+    // only safe alongside a contract-version bump and a deprecation entry, and
+    // this test is what forces that conversation. Renaming
+    // OPEN_DESIGN_IN_KANVAS -> OPEN_DESIGN_IN_EXTENSION silently broke opening a
+    // design because nothing pinned them.
+    expect(MESHERY_EXTENSION_EVENT).toEqual({
+      K8sContextsUpdated: 'K8S_CONTEXTS_UPDATED',
+      OpenViewScopedToDesign: 'OPEN_VIEW_SCOPED_TO_DESIGN',
+      OpenDesignInExtension: 'OPEN_DESIGN_IN_EXTENSION',
+      OpenViewInExtension: 'OPEN_VIEW_IN_EXTENSION',
+      MergeDesign: 'MERGE_DESIGN',
+      DispatchToMesheryStore: 'DISPATCH_TO_MESHERY_STORE',
+      FeatureRequiresUserAccount: 'FeatureRequiresUserAccount',
+      MissingPermission: 'MISSING_PERMISSION',
+      MissingCapability: 'MISSING_CAPABILITY'
+    });
+  });
+
+  it('recognises contract events and rejects unknown literals', () => {
+    expect(
+      isMesheryExtensionEvent({
+        type: MESHERY_EXTENSION_EVENT.OpenDesignInExtension,
+        data: { designId: 'd', designName: 'n' }
+      })
+    ).toBe(true);
+    // The pre-fix literal. A subscriber that receives this is talking to a host
+    // built against a different contract and must say so rather than no-op.
+    expect(isMesheryExtensionEvent({ type: 'OPEN_DESIGN_IN_KANVAS', data: {} })).toBe(false);
+  });
+
+  it('rejects non-event values without throwing', () => {
+    expect(isMesheryExtensionEvent(null)).toBe(false);
+    expect(isMesheryExtensionEvent(undefined)).toBe(false);
+    expect(isMesheryExtensionEvent('OPEN_DESIGN_IN_EXTENSION')).toBe(false);
+    expect(isMesheryExtensionEvent({})).toBe(false);
+  });
+});
+
+describe('Meshery extension contract — injected capabilities', () => {
+  it('reports a fully-satisfying host as satisfied', () => {
+    const report = reportInjectedCapabilities(satisfyingInjectProps());
+
+    expect(report.missing).toEqual([]);
+    expect(report.missingHooks).toEqual([]);
+    expect(isInjectedCapabilityReportSatisfied(report)).toBe(true);
+    expect(describeInjectedCapabilityReport(report)).toBeNull();
+  });
+
+  it('names the capability a host stopped injecting', () => {
+    // This is the capabilitiesRegistry -> providerCapabilities incident in
+    // miniature: the host drops a key, every read yields undefined, and the
+    // failure used to surface as an unrelated crash far from the cause.
+    const injectProps = satisfyingInjectProps();
+    delete (injectProps as Record<string, unknown>).providerCapabilities;
+
+    const report = reportInjectedCapabilities(injectProps);
+
+    expect(report.missing).toEqual(['providerCapabilities']);
+    expect(isInjectedCapabilityReportSatisfied(report)).toBe(false);
+    expect(describeInjectedCapabilityReport(report)).toContain('providerCapabilities');
+  });
+
+  it('names a missing nested hook', () => {
+    const injectProps = { ...satisfyingInjectProps(), hooks: { CAN: () => true } };
+
+    const report = reportInjectedCapabilities(injectProps);
+
+    expect(report.missingHooks).toEqual(['useDynamicComponent', 'useFilterK8sContexts']);
+    expect(describeInjectedCapabilityReport(report)).toContain('useFilterK8sContexts');
+  });
+
+  it('does not double-report hooks when the whole hooks bag is absent', () => {
+    const injectProps = satisfyingInjectProps();
+    delete (injectProps as Record<string, unknown>).hooks;
+
+    const report = reportInjectedCapabilities(injectProps);
+
+    expect(report.missing).toContain('hooks');
+    expect(report.missingHooks).toEqual([]);
+  });
+
+  it('flags deprecated aliases alongside their replacement without failing the check', () => {
+    const report = reportInjectedCapabilities({
+      ...satisfyingInjectProps(),
+      capabilitiesRegistry: {},
+      CapabilitiesRegistryClass: class {}
+    });
+
+    expect(report.deprecated).toEqual([
+      { provided: 'capabilitiesRegistry', replacement: 'providerCapabilities' },
+      { provided: 'CapabilitiesRegistryClass', replacement: 'ProviderUiAccessControlClass' }
+    ]);
+    // Legacy bundles still depend on these, so their presence is informational.
+    expect(isInjectedCapabilityReportSatisfied(report)).toBe(true);
+  });
+
+  it('does not report the host handshake field as an unknown key', () => {
+    // The host advertises `contractVersion` on the same bag. Reporting the
+    // host's own handshake field as unrecognized is noise, and noise is what
+    // trains people to stop reading the report.
+    const report = reportInjectedCapabilities({
+      ...satisfyingInjectProps(),
+      contractVersion: MESHERY_EXTENSION_CONTRACT_VERSION
+    });
+
+    expect(report.unrecognized).toEqual([]);
+  });
+
+  it('reports keys this build has never heard of as unrecognized', () => {
+    const report = reportInjectedCapabilities({
+      ...satisfyingInjectProps(),
+      someCapabilityFromANewerHost: () => null
+    });
+
+    expect(report.unrecognized).toEqual(['someCapabilityFromANewerHost']);
+    expect(isInjectedCapabilityReportSatisfied(report)).toBe(true);
+  });
+
+  it('treats a null or undefined bag as every capability missing', () => {
+    for (const bag of [null, undefined]) {
+      const report = reportInjectedCapabilities(bag);
+      expect(report.missing).toEqual([...MESHERY_EXTENSION_CAPABILITIES]);
+      expect(describeInjectedCapabilityReport(report)).toContain('contract');
+    }
+  });
+
+  it('keeps every deprecated alias pointing at a real capability or an outright removal', () => {
+    for (const replacement of Object.values(MESHERY_EXTENSION_DEPRECATED_CAPABILITIES)) {
+      if (replacement === null) continue;
+      expect(MESHERY_EXTENSION_CAPABILITIES).toContain(replacement);
+    }
+  });
+
+  it('never lists a deprecated alias as a current capability', () => {
+    for (const alias of Object.keys(MESHERY_EXTENSION_DEPRECATED_CAPABILITIES)) {
+      expect(MESHERY_EXTENSION_CAPABILITIES).not.toContain(alias);
+    }
+  });
+});
+
+describe('Meshery extension contract — version', () => {
+  it('is a positive integer hosts and extensions can compare', () => {
+    expect(Number.isInteger(MESHERY_EXTENSION_CONTRACT_VERSION)).toBe(true);
+    expect(MESHERY_EXTENSION_CONTRACT_VERSION).toBeGreaterThan(0);
+  });
+});
--- a/src/__testing__/mesheryExtensionContract.test.ts
+++ b/src/__testing__/mesheryExtensionContract.test.ts
@@ -63,6 +63,64 @@ describe('Meshery extension contract — event literals', () => {
     expect(isMesheryExtensionEvent('OPEN_DESIGN_IN_EXTENSION')).toBe(false);
     expect(isMesheryExtensionEvent({})).toBe(false);
   });
+
+  it('rejects a known literal carrying a payload this build cannot read', () => {
+    // The rename half of the same skew: the host keeps publishing a literal we
+    // recognise but renames the field inside it, so every subscriber reads
+    // undefined. Narrowing on the discriminant alone would call this valid and
+    // hand the subscriber a payload that throws on first property access.
+    const type = MESHERY_EXTENSION_EVENT.OpenDesignInExtension;
+
+    expect(isMesheryExtensionEvent({ type, data: null })).toBe(false);
+    expect(isMesheryExtensionEvent({ type })).toBe(false);
+    expect(isMesheryExtensionEvent({ type, data: {} })).toBe(false);
+    // `designId` renamed to `id` - recognised literal, unreadable payload.
+    expect(isMesheryExtensionEvent({ type, data: { id: 'd', designName: 'n' } })).toBe(false);
+    expect(isMesheryExtensionEvent({ type, data: { designId: 1, designName: 'n' } })).toBe(false);
+  });
+
+  it('accepts a valid payload for every event literal', () => {
+    const valid: Record<(typeof MESHERY_EXTENSION_EVENT_TYPES)[number], unknown> = {
+      K8S_CONTEXTS_UPDATED: { selectedK8sContexts: ['ctx'] },
+      OPEN_VIEW_SCOPED_TO_DESIGN: { designId: 'd', designName: 'n' },
+      OPEN_DESIGN_IN_EXTENSION: { designId: 'd', designName: 'n' },
+      OPEN_VIEW_IN_EXTENSION: { viewId: 'v', viewName: 'n' },
+      MERGE_DESIGN: { id: 'i', name: 'n' },
+      DISPATCH_TO_MESHERY_STORE: { type: 'redux/action' },
+      FeatureRequiresUserAccount: { feature: 'f' },
+      MISSING_PERMISSION: { keyId: 'k' },
+      MISSING_CAPABILITY: { capabilityId: 'c' }
+    };
+
+    for (const type of MESHERY_EXTENSION_EVENT_TYPES) {
+      expect(isMesheryExtensionEvent({ type, data: valid[type] })).toBe(true);
+    }
+  });
+
+  it('tolerates extra payload fields so additive changes stay compatible', () => {
+    // Only a rename or removal is a contract break. A newer host adding a field
+    // to a payload must not make an older bundle reject the whole event.
+    expect(
+      isMesheryExtensionEvent({
+        type: MESHERY_EXTENSION_EVENT.MergeDesign,
+        data: { id: 'i', name: 'n', addedByANewerHost: true }
+      })
+    ).toBe(true);
+    expect(
+      isMesheryExtensionEvent({
+        type: MESHERY_EXTENSION_EVENT.DispatchToMesheryStore,
+        data: { type: 'redux/action', payload: { a: 1 } }
+      })
+    ).toBe(true);
+  });
+
+  it('does not mistake an inherited Object member for a payload check', () => {
+    // The check table is a plain object, so an event whose `type` names something
+    // on Object.prototype must not resolve to an inherited function and pass.
+    expect(isMesheryExtensionEvent({ type: 'toString', data: {} })).toBe(false);
+    expect(isMesheryExtensionEvent({ type: 'constructor', data: {} })).toBe(false);
+    expect(isMesheryExtensionEvent({ type: 'hasOwnProperty', data: {} })).toBe(false);
+  });
 });
 
 describe('Meshery extension contract — injected capabilities', () => {
@@ -171,5 +229,60 @@ describe('Meshery extension contract — version', () => {
   it('is a positive integer hosts and extensions can compare', () => {
     expect(Number.isInteger(MESHERY_EXTENSION_CONTRACT_VERSION)).toBe(true);
     expect(MESHERY_EXTENSION_CONTRACT_VERSION).toBeGreaterThan(0);
+  });
+
+  it('treats a matching advertised version as compatible', () => {
+    const report = reportInjectedCapabilities({
+      ...satisfyingInjectProps(),
+      contractVersion: MESHERY_EXTENSION_CONTRACT_VERSION
+    });
+
+    expect(report.contractVersionMismatch).toBeNull();
+    expect(isInjectedCapabilityReportSatisfied(report)).toBe(true);
+  });
+
+  it('treats a host that advertises no version as the supported legacy case', () => {
+    // Every host deployed before this handshake existed. They predate the field
+    // rather than disagreeing about it, so flagging them would make the report
+    // fire everywhere at once and teach maintainers to ignore it.
+    const report = reportInjectedCapabilities(satisfyingInjectProps());
+
+    expect(report.contractVersionMismatch).toBeNull();
+    expect(isInjectedCapabilityReportSatisfied(report)).toBe(true);
+    expect(describeInjectedCapabilityReport(report)).toBeNull();
+  });
+
+  it('fails the handshake loudly when the advertised version differs', () => {
+    const report = reportInjectedCapabilities({
+      ...satisfyingInjectProps(),
+      contractVersion: MESHERY_EXTENSION_CONTRACT_VERSION + 1
+    });
+
+    expect(report.contractVersionMismatch).toEqual({
+      host: MESHERY_EXTENSION_CONTRACT_VERSION + 1,
+      extension: MESHERY_EXTENSION_CONTRACT_VERSION
+    });
+    // Every capability is present, so nothing else in the report would catch this.
+    expect(report.missing).toEqual([]);
+    expect(isInjectedCapabilityReportSatisfied(report)).toBe(false);
+
+    const message = describeInjectedCapabilityReport(report);
+    expect(message).toContain(`v${MESHERY_EXTENSION_CONTRACT_VERSION + 1}`);
+    expect(message).toContain(`v${MESHERY_EXTENSION_CONTRACT_VERSION}`);
+  });
+
+  it('treats a non-numeric advertised version as a mismatch rather than trusting it', () => {
+    // `injectProps` is untrusted input; a version we cannot compare is not a
+    // version we may assume matches.
+    const report = reportInjectedCapabilities({
+      ...satisfyingInjectProps(),
+      contractVersion: `${MESHERY_EXTENSION_CONTRACT_VERSION}`
+    });
+
+    expect(report.contractVersionMismatch).toEqual({
+      host: `${MESHERY_EXTENSION_CONTRACT_VERSION}`,
+      extension: MESHERY_EXTENSION_CONTRACT_VERSION
+    });
+    expect(isInjectedCapabilityReportSatisfied(report)).toBe(false);
   });
 });
--- a/src/actors/mesheryExtensionContract.ts
+++ b/src/actors/mesheryExtensionContract.ts
@@ -393,9 +393,12 @@ export const describeInjectedCapabilityReport = (
   if (report.missingHooks.length > 0) {
     parts.push(`missing hooks: ${report.missingHooks.join(', ')}`);
   }
+  // Deliberately no closing claim about *why* they disagree. When a version
+  // mismatch is known it is already named above; when it is not, the host and
+  // this bundle disagree while advertising the same version, and asserting a
+  // version skew would send the reader to compare two numbers that match.
   return (
     `Meshery host did not satisfy the extension contract (v${MESHERY_EXTENSION_CONTRACT_VERSION}): ` +
-    `${parts.join('; ')}. Host and extension bundle were built against different ` +
-    'contract revisions; the features these keys back will not work.'
+    `${parts.join('; ')}. The features these keys back will not work.`
   );
 };
--- a/src/actors/mesheryExtensionContract.ts
+++ b/src/actors/mesheryExtensionContract.ts
@@ -156,18 +156,60 @@ export const MESHERY_EXTENSION_EVENT_TYPES = Object.values(
   MESHERY_EXTENSION_EVENT
 ) as readonly MesheryExtensionEventType[];
 
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null;
+
+const isString = (value: unknown): value is string => typeof value === 'string';
+
+const isStringArray = (value: unknown): value is string[] =>
+  Array.isArray(value) && value.every(isString);
+
 /**
- * Narrows an event received from the bus to a known contract member.
+ * Payload check per event literal, keyed by the literal itself.
+ *
+ * The type discriminant alone is not the contract - the payload shape is half of
+ * it. A host that renames `designId` to `id` keeps publishing a literal this
+ * build recognises while every subscriber reads `undefined`, which is the same
+ * silent no-op the literal map exists to prevent. Checking only required fields
+ * (never rejecting extra ones) keeps additive payload changes compatible while
+ * making a rename or removal visible.
+ *
+ * `satisfies Record<MesheryExtensionEventType, ...>` is the completeness guard:
+ * adding a member to `MesheryExtensionEvent` without a check here fails to
+ * compile, so this table cannot drift behind the union.
+ */
+const MESHERY_EXTENSION_EVENT_DATA_CHECKS = {
+  K8S_CONTEXTS_UPDATED: data => isStringArray(data.selectedK8sContexts),
+  OPEN_VIEW_SCOPED_TO_DESIGN: data => isString(data.designId) && isString(data.designName),
+  OPEN_DESIGN_IN_EXTENSION: data => isString(data.designId) && isString(data.designName),
+  OPEN_VIEW_IN_EXTENSION: data => isString(data.viewId) && isString(data.viewName),
+  MERGE_DESIGN: data => isString(data.id) && isString(data.name),
+  DISPATCH_TO_MESHERY_STORE: data =>
+    isString(data.type) && (data.payload === undefined || isRecord(data.payload)),
+  FeatureRequiresUserAccount: data => isString(data.feature),
+  MISSING_PERMISSION: data => isString(data.keyId),
+  MISSING_CAPABILITY: data => isString(data.capabilityId)
+} as const satisfies Record<MesheryExtensionEventType, (data: Record<string, unknown>) => boolean>;
+
+/**
+ * Narrows an event received from the bus to a known contract member, checking
+ * both the type discriminant and the payload it promises to carry.
  *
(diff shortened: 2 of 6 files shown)


pontusringblom (approved):

pontusringblom:
@rishiraj38, please note the changes pertaining to permissions.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+ */
+import type { EventBus } from './eventBus';
+
+/**
+ * Incremented only when an event or capability is renamed or removed. The host
+ * injects this as `contractVersion`; extensions compare it against the value they
+ * were built against and surface a loud, actionable mismatch instead of failing
+ * feature-by-feature at runtime.
+ */
+export const MESHERY_EXTENSION_CONTRACT_VERSION = 1;
```
marblom007, line 44: **Alignment question: this contract is declared in the design system, not in `meshery/schemas`.**

This is the core of the review question. The module is a genuine improvement over hand-duplicated literals, but it establishes sistent as the source of truth for a **cross-repo service contract**, which cuts against the stated architecture:

- `meshery/schemas` already owns permission keys end to end: `build/permissions.csv` → `build/generate-permissions-ts.js` → `typescript/permissions.ts` (+ `models/permissions/permissions.go`), fingerprinted by `PERMISSIONS_INDEX_ID` and versioned in `permissions_history/`. That is a generated, hashed, multi-language artifact.
- It also already has a `capability` construct (`schemas/constructs/v1beta1/capability/capability.yaml`, with generated Go and TS).
- Sistent's role in that split is the design system - UX principles and UI components - and it already models this correctly elsewhere: every file under `src/schemas/` is a one-line re-export with a "source-of-truth" comment (`export { ModelImportRjsfSchemaV1Beta2 as default } from '@meshery/schemas'`), and `permissions.tsx` itself sources `Key` from `@meshery/schemas/permissions`.

`MESHERY_EXTENSION_CONTRACT_VERSION` is a hand-incremented integer in a TS file; `MESHERY_EXTENSION_CAPABILITIES` is a hand-maintained array. I diffed the array against Meshery's actual `injectProps` in `ui/components/layout/Navigator/NavigatorExtension.tsx` and it is **exactly right today** (all 30 keys, plus `resolver` and `CapabilitiesRegistryClass` correctly in the deprecated table). But nothing keeps it right: the truth lives in a `useMemo` in another repo, and drift in the direction that matters most - Meshery adds a capability - is silent here.

**Concrete direction**, which would let this same module survive unchanged as a re-export:
1. Land the event/capability contract as a construct in `meshery/schemas` (`constructs/v1beta1/extension-contract/`), generating TS **and** Go so Meshery Server and Cloud can validate the same names.
2. Have the generator emit `MESHERY_EXTENSION_CAPABILITIES` / `MESHERY_EXTENSION_EVENT` and the contract version, so a bump is a schema change with a history entry rather than an edit anyone can forget.
3. Reduce this file to `export { ... } from '@meshery/schemas/extensionContract'`, matching the `src/schemas/` pattern already established in this repo.

Not blocking the PR - the compile-time safety here is real and worth having now. But it should land as a follow-up before a second consumer starts deriving from sistent, because the migration cost grows with every repo that imports from here instead of from schemas.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+};
+
+/* -------------------------------------------------------------------------- */
+/* Events: published by sistent's PermissionShield onto whichever bus it is     */
+/* handed, which in Meshery is the host <-> extension bus.                      */
+/* -------------------------------------------------------------------------- */
+
+export type MissingPermissionReason = {
+  type: 'MISSING_PERMISSION';
+  data: { keyId: string };
```
marblom007, line 96: **`keyId` is typed `string`, dropping the branded `PermissionKey` that `meshery/schemas` exists to provide.**

`@meshery/schemas/permissions` defines `PermissionKey` as `string & { readonly __brand: 'PermissionKey' }` precisely so a permission identifier cannot be confused with an arbitrary string, and `Key.id` is typed as it. This file - which is where the permission event is now declared - flattens it back to `string`.

Failure it allows: a publisher emits `{ type: 'MISSING_PERMISSION', data: { keyId: 'create-design' } }` (a human-readable slug rather than the key UUID). It compiles on both sides. The subscriber in Meshery/Kanvas looks it up in the generated `Keys` table, finds nothing, and renders a shield with no key metadata - the same silent-no-op class of bug this module was written to eliminate, one level down.

Directly relevant to the schemas-as-single-source-of-truth objective: an event payload that carries a permission key should be typed by the package that generates permission keys.

**Not auto-fixed** - typing it strictly is a small cascade, not a one-liner, and it changes an exported type:
```ts
import type { PermissionKey } from '@meshery/schemas/permissions';
export type MissingPermissionReason = {
  type: 'MISSING_PERMISSION';
  data: { keyId: PermissionKey };
};
```
`createCanShow` currently builds `keyId` from `Key?.id || Key?.action || ''`, which widens to `string`, so the legacy `action`/`subject` fallbacks in `HasKeyProps` need to be retired (or narrowed) in the same change. Worth doing deliberately rather than as a review edit.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+
+/**
+ * Narrows an event received from the bus to a known contract member.
+ *
+ * Subscribers must treat a `false` result as a defect to report, not as something
+ * to ignore: it means the publisher is emitting a literal this build has never
+ * heard of, which is exactly the host/extension skew that used to present as a
+ * silently dead feature.
+ */
+export const isMesheryExtensionEvent = (event: unknown): event is MesheryExtensionEvent =>
```
marblom007: **`isMesheryExtensionEvent` narrows to `MesheryExtensionEvent` while only checking `type` - the `data` guarantee is fabricated.**

Every member of the union carries a `data` object, so narrowing to the union promises the caller that `event.data` exists. The predicate never looks at it.

Failure: a host built against a newer contract publishes `{ type: 'MERGE_DESIGN' }` with a renamed payload field, or any malformed event with a valid `type`. The guard returns `true`, TypeScript narrows to `MergeDesignEvent`, and the subscriber's `event.data.id` throws `TypeError: Cannot read properties of undefined`. That is strictly worse than no guard: the crash now happens *inside* the branch the developer believed was validated, and the doc above tells subscribers to treat `false` as the defect signal - so the one case that does reach them unvalidated is the one they were told was safe.

**Fixed** - the predicate now also requires `data` to be a non-null object, with tests for `{ type: MERGE_DESIGN }` and `data: null`. Note this is still a shallow check; per-member payload validation would need a schema, which is another argument for generating this contract from `meshery/schemas`.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+  };
+};
+
+/**
+ * True when the host satisfied every declared capability. Deprecated and
+ * unrecognized keys are informational and do not fail the check — only an
+ * outright missing capability is a broken contract.
+ */
+export const isInjectedCapabilityReportSatisfied = (report: InjectedCapabilityReport): boolean =>
+  report.missing.length === 0 && report.missingHooks.length === 0;
```
marblom007: **Confirming @coderabbitai's finding, and it runs deeper than the version comparison.**

`contractVersion` is declared in `MESHERY_EXTENSION_HANDSHAKE_FIELDS`, excluded from `unrecognized`, and then never read. So the module's own promise - "extensions compare it against the value they were built against and surface a loud, actionable mismatch" - is not implemented anywhere: a host advertising `contractVersion: 7` produces a fully "satisfied" report and `describeInjectedCapabilityReport()` returns `null`.

The second half is more damaging. The header says additive changes need **no version bump**, but `reportInjectedCapabilities` reports a newly-added capability as `missing`, `isInjectedCapabilityReportSatisfied` returns `false`, and the message then asserts:

> The host and this extension bundle were built against different contract versions

In the additive case both sides report the *same* version. The most common real failure sends the maintainer to diff two identical numbers, and the message is confidently wrong about the cause.

**Fixed:**
- `InjectedCapabilityReport` gains `hostContractVersion` and `contractVersionMismatch: { host, extension } | null`.
- Absent version = the legacy case (hosts predating the handshake, which today is all of them - Meshery's `NavigatorExtension` injects no `contractVersion`) and is not faulted.
- A differing version, or a non-numeric one, is a mismatch and fails `isInjectedCapabilityReportSatisfied`.
- `describeInjectedCapabilityReport` now states the cause it actually knows: "renamed or removed between those versions" on a real mismatch, "additive key the host has not shipped yet" when the versions agree.

Covered by five new tests: matching, absent/legacy, mismatched, malformed (`'v1'`), and the same-version-but-missing-key case that used to be misreported.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+ */
+export const MESHERY_EXTENSION_EVENT = {
+  K8sContextsUpdated: 'K8S_CONTEXTS_UPDATED',
+  OpenViewScopedToDesign: 'OPEN_VIEW_SCOPED_TO_DESIGN',
+  OpenDesignInExtension: 'OPEN_DESIGN_IN_EXTENSION',
+  OpenViewInExtension: 'OPEN_VIEW_IN_EXTENSION',
+  MergeDesign: 'MERGE_DESIGN',
+  DispatchToMesheryStore: 'DISPATCH_TO_MESHERY_STORE',
+  FeatureRequiresUserAccount: 'FeatureRequiresUserAccount',
+  MissingPermission: 'MISSING_PERMISSION',
```
marblom007, line 136: **Sistent's own publish site still used the raw string literal this map exists to replace.**

`MESHERY_EXTENSION_EVENT` is introduced so "a rename here breaks compilation at every call site in every repo" - but the single place sistent publishes onto this bus, `createCanShow` in `src/custom/permissions.tsx`, still built the event with an inline literal:

```ts
const reason = predicateRes?.[1] || {
  type: 'MISSING_PERMISSION',   // <- not derived from the map
  data: { keyId: actionString }
};
```

Renaming `MissingPermission` in the map would leave this publishing the old literal, compiling cleanly, and every subscriber's `event.type === MESHERY_EXTENSION_EVENT.MissingPermission` would quietly stop matching - the exact `OPEN_DESIGN_IN_KANVAS` failure mode, reintroduced by the module meant to prevent it. The object literal was also untyped, so nothing forced it to be a `ReasonEvent` at all.

**Fixed:** annotated `const reason: ReasonEvent` and switched to `MESHERY_EXTENSION_EVENT.MissingPermission`.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * Compile-time completeness guard. Adding a member to `MesheryExtensionEvent`
+ * without adding it to `MESHERY_EXTENSION_EVENT` makes this assignment fail, so
+ * the map can never drift behind the union it describes.
+ */
+type UnmappedEventType = Exclude<
+  MesheryExtensionEventType,
+  (typeof MESHERY_EXTENSION_EVENT)[keyof typeof MESHERY_EXTENSION_EVENT]
+>;
+const _everyEventIsMapped: UnmappedEventType extends never ? true : never = true;
```
marblom007, line 151: **The completeness guard ships dead code to every consumer, and does not actually assert what it looks like it asserts.**

```ts
const _everyEventIsMapped: UnmappedEventType extends never ? true : never = true;
void _everyEventIsMapped;
```

Two problems:

1. **It emits at runtime.** A `const` plus a `void` expression survive into `dist/index.js` and `dist/index.mjs` for a check the compiler finished before emit. Type-level assertions should cost nothing.
2. **The conditional form is unsound as a guard.** `never` is assignable to every type, so `X extends never ? true : never` yields `never` on failure - and `const x: never = true` does error, so this one happens to work; but the idiom is a foot-gun that silently passes in the very similar `Expect<T extends true>` shape people reach for next.

**Fixed** with a type-only assertion that fails on the constraint instead:
```ts
type AssertNever<T extends never> = T;
export type EveryEventIsMapped = AssertNever<UnmappedEventType>;
```
Adding a member to `MesheryExtensionEvent` without adding it to `MESHERY_EXTENSION_EVENT` now fails to compile, with zero runtime output.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+
+export type MesheryExtensionHook = (typeof MESHERY_EXTENSION_HOOKS)[number];
+
+/**
+ * Capability keys the host still injects purely so extension bundles published
+ * against an older contract keep mounting. New code must read the replacement.
+ * The host is expected to keep supplying these until the corresponding bundles
+ * age out; dropping one is a breaking contract change.
+ */
+export const MESHERY_EXTENSION_DEPRECATED_CAPABILITIES = {
```
marblom007, line 274: **The deprecation table's doc comment is factually wrong about the current host, which will mislead the next person to read it.**

> The host is expected to keep supplying these until the corresponding bundles age out; dropping one is a breaking contract change.

Checked against Meshery's actual injection site (`ui/components/layout/Navigator/NavigatorExtension.tsx`): the host supplies `CapabilitiesRegistryClass` and `resolver`, but it does **not** supply `capabilitiesRegistry` - that key is already gone. So the comment asserts an invariant the host is currently violating, which is how a maintainer ends up "fixing" Meshery to re-add a key nobody wants back.

**Fixed** - the comment now describes what the table is *for* (naming the replacement when a host does still supply a retired key, and preventing a retired key from being re-added to `MESHERY_EXTENSION_CAPABILITIES`) and notes explicitly which of the three the host still injects.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+    // reporting all three hooks again would be noise, so only look inside when
+    // the bag is actually there.
+    missingHooks:
+      provided.hooks === undefined
+        ? []
+        : MESHERY_EXTENSION_HOOKS.filter(hook => hooks[hook] === undefined),
+    deprecated: Object.entries(MESHERY_EXTENSION_DEPRECATED_CAPABILITIES)
+      .filter(([key]) => provided[key] !== undefined)
+      .map(([provided_, replacement]) => ({ provided: provided_, replacement })),
+    unrecognized: providedKeys.filter(key => !known.has(key))
```
marblom007, line 358: **Drift detection is asymmetric: unknown top-level keys are reported, unknown nested hooks are not.**

`unrecognized` is computed from `Object.keys(provided)`, so a newer host adding `injectProps.someNewThing` is surfaced. But three of the host's capabilities live inside `injectProps.hooks` (`CAN`, `useDynamicComponent`, `useFilterK8sContexts`), and a newer host adding `hooks.useSomethingNew` produces an empty report.

That is the wrong way round for the stated purpose. The doc says an unrecognized key "is the signal that this bundle predates the host and may be missing behaviour the host expects it to have" - and a new *hook* is exactly the kind of thing a host adds and then starts depending on. The report goes quiet precisely where it is most needed.

**Fixed** - added `unrecognizedHooks` to `InjectedCapabilityReport`, populated only when the `hooks` bag is present (mirroring the existing, correct `missingHooks` behaviour so an absent bag is not double-reported). Informational, like `unrecognized`, so it does not fail the satisfaction check. Two tests added.

src/base/DateTimePicker/DateTimePicker.tsx
```diff
@@ -1,19 +1,47 @@
...
-const DateTimePicker = React.forwardRef<HTMLDivElement, MuiDateTimePickerProps>(
-  (props, ref) => {
-    return (
+/**
+ * `@mui/x-date-pickers` is declared an OPTIONAL peer dependency, and most
+ * consumers never render a date picker, so most consumers do not install it.
+ * Importing it at module scope contradicted that: this module is reached from the
+ * package barrel, so merely writing `import { anything } from '@sistent/sistent'`
+ * executed a top-level require of the optional package and threw
+ * `Cannot find module '@mui/x-date-pickers/AdapterDateFns'` for anyone without
```
marblom007, line 10: **This fix is correct, but it is only half the problem - `date-fns` is also an optional peer and is still imported eagerly.**

The diagnosis in this comment is exactly right, and I verified the `@mui/x-date-pickers` half is genuinely fixed (the built bundle now contains only `import("@mui/x-date-pickers/...")`, no top-level require). But `src/custom/UniversalFilter.tsx:3` does:

```ts
import { subDays, subMonths, subYears } from 'date-fns';
```

`date-fns` is declared optional in `peerDependenciesMeta` alongside `@mui/x-date-pickers`, and `UniversalFilter` is reached from the barrel via `export * from './custom'`. So the failure described here survives, through the other peer. Verified against the artifacts built from this branch:

```
$ grep -c 'require("date-fns")' dist/index.js   # 1   (top-level, CJS)
$ grep -c 'from"date-fns"'      dist/index.mjs  # 1   (top-level, ESM)
```

and end-to-end, blocking both optional peers from module resolution and requiring the built barrel:

```
PR HEAD:     FAIL barrel threw: Cannot find module 'date-fns'
with fix:    OK  barrel imported without the optional peers
```

So on this branch `import { Button } from '@sistent/sistent'` still dies before anything renders, in exactly the Kanvas clean-install scenario described above.

**Fixed** - the three calls are trivial calendar arithmetic, so they now come from a new dependency-free leaf module `src/utils/date.utils.ts`, matching `date-fns` semantics including end-of-month clamping (`subtractMonths(May 31, 3)` is Feb 28, not Mar 3 - the naive `setMonth()` gets this wrong and would silently widen a "last 3 months" filter by three days). That removes `date-fns` from the runtime graph entirely rather than deferring it, which is better than lazy-loading here since the arithmetic is ~15 lines.

Guarded against recurrence by `src/__testing__/optionalPeerDependencies.test.ts`, which walks every non-test module under `src/` and fails on any eager import of either optional peer. Verified it distinguishes the bug from the fix: it flags `import ... from 'date-fns'` and `require('date-fns')`, and allows `await import(...)` and type-only imports. It would have caught both halves of this.

src/base/DateTimePicker/DateTimePicker.tsx
```diff
@@ -1,19 +1,47 @@
...
+ * optional: the cost is paid only by code that actually renders one, which is
+ * what `peerDependenciesMeta.optional` already advertises.
+ */
+const LazyDateTimePicker = React.lazy(async () => {
+  const [{ AdapterDateFns }, { LocalizationProvider }, { DateTimePicker: MuiDateTimePicker }] =
+    await Promise.all([
+      import('@mui/x-date-pickers/AdapterDateFns'),
+      import('@mui/x-date-pickers/LocalizationProvider'),
+      import('@mui/x-date-pickers/DateTimePicker')
+    ]);
```
marblom007, line 24: **Deferring the import also defers the failure, into a place with no actionable message.**

The stated motivation is that the old error pointed at sistent rather than at the missing peer. But moving it into `React.lazy` does not fix that - it makes it worse in one respect: the module-not-found now surfaces mid-render from inside a rejected promise, propagating to whatever error boundary happens to be above `<UniversalFilter>`, with no import statement anywhere near the stack. A consumer who genuinely did not install the optional peer gets an async component crash instead of a load-time error, and still no sentence telling them what to install.

**Fixed** - the `Promise.all` now has a `.catch` that rethrows with the remediation, preserving the original via `cause`:

```
<DateTimePicker> requires the optional peer dependencies @mui/x-date-pickers and date-fns.
Install them, or do not render <DateTimePicker> / <UniversalFilter dateRange>.
```

Two things worth flagging that I did **not** change, since both are judgement calls for you:

1. **`fallback={null}` in a form field.** The reasoning in the comment is sound for a warm chunk, but on a cold first open of the `UniversalFilter` date range the two pickers pop in after a network round-trip, shifting the panel. Worth a fixed-height placeholder rather than `null` if you have seen it flash.
2. **Dynamic `import()` in the CJS build.** esbuild emits a native `import()` inside `dist/index.js`. Node handles it, bundlers handle it, but Jest's default (non-`--experimental-vm-modules`) runtime cannot execute a dynamic import in a CJS module - so a *consumer* test that actually renders a `DateTimePicker` may now fail where it previously passed. Since the motivating bug was Kanvas Jest suites, that trade is worth confirming against a real Kanvas run before merge.

src/custom/permissions.tsx
```diff
@@ -16,19 +20,12 @@ const DIVIDER_SX = {
...
-  data: {
-    capabilityId: string;
-  };
-};
+// These two reason events are published onto whichever bus the caller supplies,
+// which in Meshery is the host <-> extension bus. They are therefore part of the
+// extension contract and are declared there, so the host's typed bus and every
+// extension's subscriber derive them from one place. Re-exported here to keep the
+// long-standing `@sistent/sistent` import paths working.
+export type { MissingCapabilityReason, MissingPermissionReason };
```
marblom007, line 28: **Adopting the contract's prescribed bus type breaks `createCanShow` at compile time - the one integration point this PR calls out.**

The module doc instructs every host to declare its bus as `EventBus<MesheryExtensionEvent>`, and `src/index.tsx` now exports `createCanShow` explicitly so that "its `eventBus` argument stops being variance-checked - the one place a host hands its event bus to sistent."

Once it *is* variance-checked, it does not compile. `EventBus<T>` is invariant in `T` (it accepts `T` in `publish` and yields `T` from `on`/`onAny`), so `EventBus<MesheryExtensionEvent>` is not assignable to `EventBus<ReasonEvent>`. Reproduced with `tsc` against this branch:

```
error TS2322: Type 'EventBus<MesheryExtensionEvent>' is not assignable to type 'EventBus<ReasonEvent>'.
  Types of parameters 'value' and 'value' are incompatible.
    Type 'MesheryExtensionEvent' is not assignable to type 'ReasonEvent'.
      Type 'K8sContextsUpdatedEvent' is not assignable to type 'MissingCapabilityReason'.
```

So the moment Meshery follows the instruction and types `mesheryEventBus`, `createCanShow(getRegistry, CAN, () => mesheryEventBus)` fails to build. Today it only works because the bare `new EventBus()` collapses to its constraint - i.e. it compiles *because* of the bug this PR is fixing. This would have landed as a green sistent CI and a red Meshery build.

**Fixed** - `createCanShow` now takes the narrowest thing it actually uses, which is variance-correct by construction:

```ts
export type ReasonEventPublisher = { publish: (event: ReasonEvent) => void };
```

A publisher that accepts the whole contract union trivially accepts a reason event, so `EventBus<MesheryExtensionEvent>` satisfies it. Verified the full host call site compiles after the change. `ReasonEventPublisher` is exported from the barrel.

src/index.tsx
```diff
@@ -51,6 +51,12 @@ export {
   type DataTableToolbarProps
 } from './custom/DataTableToolbar';
 
+// Same nested-barrel dts-drop quirk as FeedbackButton above. `createCanShow` is
+// worse than a missing type: consumers still resolve it at runtime, so the import
+// silently degrades to `any` and its `eventBus` argument stops being
+// variance-checked - the one place a host hands its event bus to sistent.
+export { createCanShow } from './custom/permissions';
```
marblom007, line 58: **`createCanShow` is re-exported without its parameter and return types, so the `any` this comment is fixing comes straight back.**

The reasoning is right, but the fix is incomplete: `createCanShow` returns a component taking `HasKeyProps<ReasonEvent>`, and takes an event-bus accessor. None of `HasKeyProps`, `ReasonEvent`, or `InvertAction` are re-exported here, and they reach the entry only through the nested `./custom` barrel - the same `rollup-plugin-dts` drop this block exists to work around. A consumer that wants to wrap the returned component, or type an `invert_action` array, cannot name any of those types and falls back to `any` at the boundary.

**Fixed:**
```ts
export {
  createCanShow,
  type HasKeyProps,
  type InvertAction,
  type ReasonEvent,
  type ReasonEventPublisher
} from './custom/permissions';
```

Confirmed present in the built `dist/index.d.ts`.

Broader point, since this is now the fifth copy of this comment block in one file: the workaround is being applied per-symbol as each one is discovered by a downstream consumer, which means the next one is also found in production. Worth either pinning the `rollup-plugin-dts` bug and fixing it once, or switching the declaration build to a mode that does not drop nested re-exports - the current approach guarantees the failure keeps recurring.

src/actors/mesheryExtensionContract.ts
```diff
@@ -0,0 +1,331 @@
...
+ *
+ * Subscribers must treat a `false` result as a defect to report, not as something
+ * to ignore: it means the publisher is emitting a literal this build has never
+ * heard of, which is exactly the host/extension skew that used to present as a
+ * silently dead feature.
+ */
+export const isMesheryExtensionEvent = (event: unknown): event is MesheryExtensionEvent =>
+  typeof event === 'object' &&
+  event !== null &&
+  MESHERY_EXTENSION_EVENT_TYPES.includes((event as { type: MesheryExtensionEventType }).type);
```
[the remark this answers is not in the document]
marblom007: Valid, fixed in 5eeff36. `isMesheryExtensionEvent` now checks each event's required payload fields, not just the discriminant. The check table is keyed by event literal and declared `satisfies Record<MesheryExtensionEventType, ...>`, so a new union member without a payload check fails to compile - the same completeness guard the literal map already had. Only required fields are checked and extra fields are tolerated, so additive payload changes stay compatible while a rename or removal is caught. Membership is tested against the literal array before indexing the table so a payload carrying `type: 'toString'` cannot resolve to an inherited Object.prototype member. Added coverage for malformed known-event payloads (null data, absent data, renamed field, wrong field type), a valid payload for every literal, extra-field tolerance, and the prototype case.

src/__testing__/mesheryExtensionContract.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  reportInjectedCapabilities
+} from '../actors/mesheryExtensionContract';
+
+/** A host bag that satisfies the contract exactly, built from the contract itself. */
+const satisfyingInjectProps = () => ({
+  ...Object.fromEntries(MESHERY_EXTENSION_CAPABILITIES.map(key => [key, () => null])),
+  hooks: Object.fromEntries(MESHERY_EXTENSION_HOOKS.map(hook => [hook, () => null]))
+});
+
+describe('Meshery extension contract — event literals', () => {
```
marblom007, line 20: **The type-only contract gets 175 lines of tests; the actual runtime bug fix in this PR gets none.**

Per the repo's own standard that tests ship with the change: this suite is genuinely good, but it exercises a module that is almost entirely compile-time. The `DateTimePicker` change is the part of this PR that fixes a real, reported, runtime failure - "in Kanvas that took out four unrelated Jest suites on a clean install" - and nothing here would catch its regression, or catch the same bug appearing in a different module.

That is not hypothetical: the same bug is live in this PR via `date-fns` in `UniversalFilter` (see the `DateTimePicker.tsx` thread). A test asserting the *property* rather than the component would have caught both.

**Added** `src/__testing__/optionalPeerDependencies.test.ts`, which walks every non-test module under `src/` and fails if any of them eagerly imports an optional peer, listing the offending paths. It correctly ignores `await import(...)` and type-only imports, and separately asserts that both peers are still marked `optional` in `package.json` so the test's premise cannot silently change.

Also added `src/__testing__/date.utils.test.ts` for the replacement date helpers, covering the end-of-month and leap-year clamping that a naive implementation gets wrong.

src/actors/mesheryExtensionContract.ts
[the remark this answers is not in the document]
marblom007: Valid, fixed in 5eeff36. The module doc promised extensions compare the advertised version and surface a loud mismatch, but nothing implemented it. `InjectedCapabilityReport` now carries `contractVersionMismatch: { host: unknown; extension: number } | null`; a differing version fails `isInjectedCapabilityReportSatisfied` and is named first in `describeInjectedCapabilityReport`. `host` is typed `unknown` because injectProps is untrusted input, so a non-numeric version is a mismatch rather than something to trust. An absent version stays the supported legacy case - every host deployed before this handshake existed predates the field rather than disagreeing about it, and flagging them all would fire the report everywhere at once and train maintainers to ignore it. Covered matching, mismatched, non-numeric, and absent.

marblom007:
@pontusringblom @rishiraj38 - to be precise about what this PR does and does not do to permissions, since #1731 touches `PermissionShield`:

**Behaviour is unchanged.** `src/custom/permissions.tsx` has no runtime change. The only edit is that the two reason-event types moved:

```
MissingPermissionReason  { type: 'MISSING_PERMISSION';  data: { keyId: string } }
MissingCapabilityReason  { type: 'MISSING_CAPABILITY'; data: { capabilityId: string } }
```

They now live in `src/actors/mesheryExtensionContract.ts` and are re-exported from `permissions.tsx`, so **every existing import path keeps working** - `import type { MissingPermissionReason } from '@sistent/sistent'` resolves exactly as before, and `ReasonEvent` is still composed from the same two.

The reason they moved: `PermissionShield` publishes them onto whichever `EventBus` the caller hands it, and in Meshery that bus *is* the host↔extension bus. That makes them part of the extension contract whether or not we declared them so. Declaring them in one place means the host's typed bus and every extension subscriber derive the literals from the same source, and a rename becomes a compile error in all three repos instead of a subscriber that quietly stops matching.

One thing worth knowing for #1731: `createCanShow` is now exported explicitly from the package entrypoint. It was previously reachable at runtime but dropped from the generated `.d.ts` by a nested-barrel quirk, so consumers silently got `any` - which meant its `eventBus` argument was not variance-checked at the one place a host hands its bus to sistent. If you are working against `createCanShow`, you will now get real types (and possibly real type errors that were being masked).

No change to `PermissionProvider`, `usePermissionUserContext`, `InvertAction`, or the shield's rendering - so #1731's positioning work should not conflict.

marblom007:
@coderabbitai review

marblom007:
## Review: alignment with the permissions refactoring and schemas-as-single-source-of-truth

Reviewed at extra-high effort. 15 findings, 13 posted inline; the two below have no anchor in the diff. `--fix` applied - what changed is listed at the end.

### Direct answer to the question

**Where it aligns.** The diagnosis is correct and valuable. Hand-duplicated string literals across sistent / meshery / meshery-extensions are exactly how `OPEN_DESIGN_IN_KANVAS` -> `OPEN_DESIGN_IN_EXTENSION` shipped broken, and typing the bus so a rename is a compile error is the right mechanism. Moving `MissingPermissionReason` / `MissingCapabilityReason` out of a UI component file and into a declared contract is a real improvement over where they were.

**Where it does not.** It makes **sistent** the source of truth for a cross-repo service contract, which is the schemas repo's job:

- `meshery/schemas` already owns permission keys end to end - `build/permissions.csv` -> `generate-permissions-ts.js` -> `typescript/permissions.ts` + `models/permissions/permissions.go`, fingerprinted by `PERMISSIONS_INDEX_ID`, versioned under `permissions_history/`. Generated, hashed, multi-language.
- It already has a `capability` construct at `schemas/constructs/v1beta1/capability/capability.yaml`.
- This PR's parallel machinery is a hand-incremented `MESHERY_EXTENSION_CONTRACT_VERSION = 1` and a hand-maintained `MESHERY_EXTENSION_CAPABILITIES` array - **no generator, no history, no Go binding**, so Meshery Server and Cloud cannot validate against the same names.
- The permission event payload is typed `keyId: string`, discarding the branded `PermissionKey` that schemas generates specifically to make that identifier unforgeable.

The tell is that sistent already models this correctly: every file under `src/schemas/` is a one-line re-export with a "source-of-truth" comment, and `permissions.tsx` itself sources `Key` from `@meshery/schemas/permissions`. This new module is the first place sistent *originates* a contract instead of re-exporting one.

I diffed `MESHERY_EXTENSION_CAPABILITIES` against Meshery's real `injectProps` in `ui/components/layout/Navigator/NavigatorExtension.tsx`: it is **exactly right today** - all 30 keys, with `resolver` and `CapabilitiesRegistryClass` correctly in the deprecated table. Nothing keeps it right. The truth is a `useMemo` in another repo, and drift in the direction that matters most - Meshery adds a capability - is silent here.

**Suggested direction** (follow-up, not a blocker): land the contract as a construct in `meshery/schemas`, generate TS **and** Go, and reduce this file to a re-export matching the `src/schemas/` pattern. Doing it before a second consumer starts importing from sistent keeps the migration cheap. Detail in the thread on `MESHERY_EXTENSION_CONTRACT_VERSION`.

---

### Two findings without an inline anchor

**1. `console.log` on every denied permission click** - `src/custom/permissions.tsx`, in `createCanShow`'s `onClick`:

```ts
console.log('cant perform action : reason', reason, eventBus);
```

Fires in production on every click of any permission-gated control the user lacks, and logs the event-bus accessor alongside it. Pre-existing, but in a file this PR touches. **Fixed** - removed; the event is published on the bus, which is the actual reporting channel.

**2. `@meshery/schemas` is a `devDependency` but appears in the published `dist/index.d.ts`.** Not fixed - it is a packaging decision, and pre-existing. `tsup` bundles the runtime via `noExternal: [/^@meshery\/schemas/]`, but the declaration bundle emits external imports:

```ts
import { Key } from '@meshery/schemas/permissions';
export { cloudApi } from '@meshery/schemas/cloudApi';
```

A consumer without `@meshery/schemas` installed gets `TS2307: Cannot find module` from sistent's own `.d.ts`. It works today only because Meshery and Kanvas happen to install it. Given that this PR deepens sistent's reliance on schemas for permission types, the dependency should be declared - `dependencies` (honest, since the runtime is bundled anyway) or `peerDependencies`. Your call.

---

### Fixes applied

| Area | Change |
|---|---|
| **Optional peers** | `date-fns` removed from the eager import graph - the barrel now loads with **both** optional peers absent (verified against the built bundle; it threw `Cannot find module 'date-fns'` before). New dependency-free `src/utils/date.utils.ts` with `date-fns`-equivalent end-of-month clamping. |
| **Host integration** | `createCanShow` takes `ReasonEventPublisher` instead of the invariant `EventBus<ReasonEvent>`, so the contract's prescribed `EventBus<MesheryExtensionEvent>` actually compiles at the call site. |
| **Contract version** | `hostContractVersion` + `contractVersionMismatch` added to the report; absent = legacy, differing/malformed = unsatisfied; the diagnostic no longer blames a version difference that does not exist. |
| **Type guard** | `isMesheryExtensionEvent` now validates `data`, so it stops narrowing payload-less events into a `TypeError`. |
| **Literals** | sistent's own publish site uses `MESHERY_EXTENSION_EVENT.MissingPermission` instead of a raw string; `const reason: ReasonEvent` annotated. |
| **Report** | `unrecognizedHooks` added so nested-bag drift is as visible as top-level drift. |
| **Emit** | Completeness guard is now type-only (`AssertNever`), no dead binding in the shipped bundle. |
| **Exports** | `HasKeyProps`, `InvertAction`, `ReasonEvent`, `ReasonEventPublisher` re-exported alongside `createCanShow`. |
| **Hygiene** | `console.log` removed; deprecation-table comment corrected against the real host. |
| **Adjacent debt** | `utils/time.utils.tsx` imported the whole `../custom` barrel, closing a `custom -> utils -> custom` cycle that dragged ESM-only `react-markdown` into any test touching a date helper. Narrowed to the leaf `../custom/CustomTooltip`. |
| **Tests** | +25. `optionalPeerDependencies.test.ts` (property-level guard against this whole bug class), `date.utils.test.ts` (clamping/leap-year), and contract tests for version matching / legacy / mismatch / malformed / same-version-missing-key, unknown hooks, and payload-less events. |

`jest` 24 suites / 428 tests pass, `eslint` clean, `tsup` build green, and the clean-install simulation passes.

### Not fixed, deliberately

- **`keyId: PermissionKey`** - correct, but it cascades into retiring the legacy `action`/`subject` fallbacks in `HasKeyProps`. Should be a deliberate change, ideally alongside the schemas migration.
- **`fallback={null}` layout shift** and **dynamic `import()` under Jest's CJS runtime** - both flagged on the `DateTimePicker` thread; the second is worth confirming against a real Kanvas run, since Kanvas Jest suites were the motivating bug.

PR #1757, opened 2026-07-28
chore(skills): consolidate agent skills into .agents/skills

sarajkrishnasingh:
## What

Makes `.agents/skills` the single source of truth for this repo's agent skills, and reduces `.claude/skills` to a relative symlink pointing at it.

## Before / after file counts

Stated so a reviewer can see nothing was dropped:

| directory | before | after |
|---|---:|---:|
| `.agents/skills` | 10 | 11 |
| `.claude/skills` | 11 | 1 (the symlink itself) |

The 10 -> 11 is exactly `cut-release/SKILL.md` arriving. All six skills - `chrome-devtools-axi`, `cut-release`, `gh-axi`, `iterate-pr`, `lavish`, `quota-axi` - resolve through both paths afterwards, and `ls .agents/skills` and `ls .claude/skills` list identically.

## Why the diff has three shapes

This repo was already *partly* consolidated, so the change is not uniform:

| skill | before | action |
|---|---|---|
| chrome-devtools-axi, gh-axi, lavish, quota-axi | per-skill symlinks `.claude/skills/X -> ../../.agents/skills/X` | superseded by one directory-level link |
| iterate-pr | real duplicate tracked in **both** directories | `.claude` copy dropped; `.agents` copy retained |
| cut-release | existed **only** in `.claude/skills` | `git mv` into `.agents/skills` |

The consolidation direction is deliberately the opposite of what the final layout suggests: content moved **from `.claude/skills` into `.agents/skills`** first, and only then did `.claude/skills` become a link. Doing it the obvious way round (`rm -rf .claude/skills && ln -s`) would have deleted tracked files.

`cut-release` moved via `git mv` and is recorded as a 100% rename, so the move is reviewable and no content changed.

## Name collisions

One name existed in both directories - `iterate-pr` - and it was **byte-identical**, verified two ways before deleting either copy: `diff -r` reported no differences, and all five tracked blob SHAs matched exactly. Per the rule that identical means keep one, the `.agents/skills` copy was retained. No other collisions, so nothing needed escalating.

## No `.codex` / `.opencode` links, and this is deliberate

Verified before creating anything, since a link nothing reads is worse than no link. Both tools **do** support skills - but both **already read `.agents/skills` natively**, which makes a link redundant rather than load-bearing:

- **Codex** - `codex-rs/core-skills/src/loader.rs` defines `AGENTS_DIR_NAME = ".agents"`, and `repo_agents_skill_roots()` walks from the project root joining `.agents/skills`.
- **OpenCode** - its skills documentation lists `.agents/skills/<name>/SKILL.md` as a project-local load path.

Claude Code is the only one of the three that does **not** discover `.agents/skills` - which is precisely why `.claude/skills` needs the symlink and the other two do not. That asymmetry is now recorded in `AGENTS.md` so the missing links do not get re-litigated.

## Two fixes beyond pure relocation

Both were regressions *this change introduced*, so they were fixed here rather than deferred:

1. **`.agents/skills/iterate-pr/SKILL.md`** - 13 references to `.claude/skills/iterate-pr/scripts/` rewritten to `.agents/skills/iterate-pr/scripts/`. A path-prefix rewrite only; no prose, flags or behaviour changed, and all three referenced scripts were verified present at the new path first. Left stale, they would have made the `.claude` compat symlink **load-bearing** for Codex and OpenCode - which otherwise never need `.claude/` at all - and would have contradicted the `AGENTS.md` rule this same branch adds.
2. **`.prettierignore`** - added `.agents/` beside the existing `.claude/`. `make lint` runs `npm run format:write`, which globs `**/*.md`, so after the move skill markdown previously shielded under `.claude/` would be rewritten in place. CI was unaffected (`format:write` precedes `format-check`); this protects local `make lint`. `eslint.config.js` and `jest.config.js` still name `.claude/` but both entries are inert (eslint globally ignores `**/*.md`; jest `roots` is `src`-only), so they were left alone rather than churned.

## Known issue this layout creates

`AGENTS.md` now documents it in full, but flagging it here too. The tracked `skills-lock.json` records four **installer-managed** skills, and commit 5482d046 shows the AXI skills installer's layout is content under `.agents/skills/<name>/` **plus** a per-skill `.claude/skills/<name>` symlink - exactly the symlinks this change removes.

On its next run the installer will try to create `.claude/skills/<name>`; because `.claude/skills` is now itself a symlink, that resolves to `.agents/skills/<name>`, which already exists as a real directory. Best case `EEXIST`. **Worst case an installer that force-replaces its destination (`rm -rf` then `symlink`) deletes the canonical skill directory and leaves a self-referential loop.** Which happens is **not established** - no installer CLI managing `skills-lock.json` is present in this repo or on `PATH`, and the destructive path was deliberately not tested by running it. Establish the installer's replace semantics before re-running it against this layout.

## Two checkout notes for reviewers

- **Existing clones may fail to check this out.** `__pycache__` is not in `.gitignore`, and the old skill text told agents to run scripts from `.claude/skills/iterate-pr/scripts/`. Anyone who did so under a Python other than 3.14 has an untracked sibling `.pyc` there. Git cannot replace a non-empty directory with a symlink, so checkout fails with `unable to create symlink .claude/skills (File exists)` and leaves a partially-updated tree. Fix: `rm -rf .claude/skills`, then check out again.
- **Windows without developer mode** (`core.symlinks=false`): `.claude/skills` materializes as a regular file, so Claude Code discovers zero project skills there. Codex and OpenCode are unaffected. Inherent to the required layout, and four of six skills were already degraded this way.

## Out of scope, reported not fixed

`.agents/skills/cut-release/SKILL.md` line 13 links `../../workflows/release-drafter.yml`, which needs three levels up, not two. This is **pre-existing and unaffected by the move** - the directory depth is identical before and after, so it was equally broken at the old path. Fixing it would mean editing skill content, which this change deliberately does not do. Worth a follow-up.

## On how this shipped

This branch was validated through the no-mistakes pipeline and **cleared every gate step it could reach** - intent, rebase, review, and the review-driven fixes above. The pipeline was then abandoned for this branch because it structurally could not finish: it cannot emit a commit carrying a `Signed-off-by` trailer, DCO is a blocking required check that validates every commit, and repairing that by amend deadlocks against the pipeline's own descendant guard. Instructing it to sign did not work either.

The final documentation commit was therefore authored directly, with a valid sign-off, rather than by the pipeline. Its content is byte-identical to what the pipeline produced (verified by blob hash). This is recorded because the reason matters - it was a tool defect, not a gap in the change or a shortcut taken for convenience.

## Summary by CodeRabbit

* **Documentation**
  * Added a release runbook covering the steps to publish npm releases, verify deployment, and coordinate follow-up updates.
  * Updated contributor guidance to clarify skill locations and release workflow references.
  * Updated pull request workflow documentation to use the current script location.
* **Chores**
  * Consolidated agent skill management under a single shared location.
  * Removed obsolete skill links and pull request workflow utilities.
  * Excluded agent guidance files from formatting checks.


suttonskate (approved):
Reviewed as a maintainer.

- Change: 15 file(s), +44/-1172, base master
- Checks at review time: 5 success

Reviewed the diff and the check results. Approving.

PR #1759, opened 2026-07-28
fix(DataTableToolbar): apply default bottom spacing

### LINKED ISSUE #1719: DataTableToolbar: bake in default bottom spacing between toolbar and table
## Current Behavior

`DataTableToolbar` renders flush against the table below it. Every consumer ends up adding the same inline override at each call site:

```tsx
<DataTableToolbar sx={{ mb: 2 }} ... />
```

In meshery-cloud this was repeated at 17 call sites (see layer5io/meshery-cloud#5740 and the review thread there); we consolidated it into a local wrapper (`ui/components/general/data-table-toolbar.tsx`) as a stopgap.

## Desired Behavior

The toolbar-to-table gap should be the component's default so every consumer inherits it without per-call-site overrides. Callers with special layouts can still override via `sx`.

Suggested: apply `marginBottom: theme.spacing(2)` in `DataTableToolbar`'s root styles, keeping `sx` merge order so caller values win.

## Follow-up

Once released, meshery-cloud can delete its wrapper and import `DataTableToolbar` from `@sistent/sistent` directly again.

khushal-winner:
Reduces marginBottom from theme.spacing(4) to theme.spacing(2) so consumers inherit the toolbar-to-table gap without per-call-site sx overrides. Callers can still override via sx.

Closes #1719

**Notes for Reviewers**

This PR fixes #

## Summary by CodeRabbit

* **Style**
  * Reduced the bottom spacing in the data table toolbar for a more compact layout.


khushal-winner:
@KhushamBansal , is it good to go?

KhushamBansal (commented):
@khushal-winner I don't think we need a test for the sx override, since inline styles already take precedence. Could you remove the additional changes in DataTableToolbar.test.tsx?

khushal-winner:
@KhushamBansal done.

KhushamBansal (approved):

PR #1774, opened 2026-08-05
feat(PermissionSessionContext): extract, export, and self-resolve from permissionKey

rishiraj38:
## Description

Extracts `PermissionSessionContext` from `PermissionShield` into a standalone, exported component so that permission failure details (required keys, function names, descriptions, and user/org/role context) can be rendered in any layout — tooltips, cards, and full-page 403 error pages — from a single source of truth.

#### Self-resolving `permissionKey` prop
`PermissionSessionContext` now accepts an optional `permissionKey` prop. When provided, the component internally resolves `displayedKeys`, `subtitle`, `categories`, and `subcategories` using the same hooks as `PermissionShield` (`useUnmetPermissionKeys`, `getPermissionKeys`, `getPermissionKeyCombinator`). This eliminates the need for callers to manually derive and pass these values.

#### `PermissionShield` refactored to delegate
`PermissionShield` previously duplicated the full key-resolution logic before passing pre-resolved props to `PermissionSessionContext`. It now passes `permissionKey` directly, keeping the **tooltip** and **card** rendering paths in sync through a single resolution path.

#### Accessibility improvements
- Copy-key control has `role="button"`, `tabIndex={0}`, and keyboard handler (`Enter`/`Space`).
- Uses promise-based `navigator.clipboard.writeText()` with `.catch()` fallback.

#### Boolean coercion fix
- Chip-container length checks wrapped with `Boolean()` to prevent React from rendering the literal string `"0"`.

#### Barrel export
- `PermissionSessionContext` and `PermissionSessionContextProps` exported from `src/index.tsx`.

### Two usage modes

| Mode | Prop | Use case |
|------|------|----------|
| **Self-resolving** | `permissionKey={Keys.X}` | Error pages, `PermissionShield` tooltip — pass the key, get the full UI |
| **Pre-resolved** | `displayedKeys={[...]}` | Advanced layouts that need custom resolution |

## Related Issues / PRs
- **Related**: [layer5io/meshery-cloud#5892](https://github.com/layer5io/meshery-cloud/pull/5892)

### Type of Change
- [x] New feature (non-breaking change which adds functionality)
- [x] Refactor / UI enhancement

https://github.com/user-attachments/assets/9dc92d8e-ab68-45a2-baa4-ebc997476ec0

## Summary by CodeRabbit

* **New Features**
  * Added a reusable permission context component with tooltip and card display options.
  * Added support for permission metadata, categories, reference links, and contextual user, organization, and role details.
  * Permission keys can now be copied using keyboard controls.
  * Exposed the permission context component and its configuration type for use in integrations.


src/custom/permissions.tsx
```diff
@@ -169,169 +227,192 @@ export const PermissionShield: React.FC<PermissionShieldProps> = ({
...
+                      color: copied ? palette.accent : isCard ? palette.muted : 'rgba(255, 255, 255, 0.7)',
+                      transition: 'color 0.2s ease',
+                      '&:hover': {
+                        color: palette.accent
+                      }
+                    }}
+                  >
+                    <KeyIcon sx={{ fontSize: `${1 * baseFontScale}rem` }} />
+                  </Box>
+                </Tooltip>
```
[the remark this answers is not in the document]
rishiraj38: Applied.

src/custom/permissions.tsx
```diff
@@ -169,169 +227,192 @@ export const PermissionShield: React.FC<PermissionShieldProps> = ({
...
-            />
-          ))}
-        </Box>
-        <Link
-          href="https://docs.meshery.io/reference/references/permissions/"
-          target="_blank"
-          rel="noopener noreferrer"
-          onClick={(e: React.MouseEvent) => e.stopPropagation()}
+      {/* Bottom row: Category/Subcategory chips + Key Reference link */}
+      {hasKeys && (categories?.length || subcategories?.length) && (
```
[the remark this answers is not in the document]
rishiraj38: Applied.

src/custom/permissions.tsx
```diff
@@ -393,6 +523,89 @@ export const PermissionShield: React.FC<PermissionShieldProps> = ({
...
+  const subcategories = uniqueDefined(displayedKeys.map((key) => key.subcategory));
+
+  const tooltipTitle = (
+    <PermissionSessionContext
+      variant="tooltip"
+      displayedKeys={displayedKeys}
+      subtitle={subtitle}
+      categories={categories}
+      subcategories={subcategories}
+    />
```
chaitanyamedidar, line 589: Now that `PermissionSessionContext` can self-resolve from `permissionKey`, should `PermissionShield` pass `permissionKey` here instead of re-deriving `displayedKeys` / `subtitle` / categories so the tooltip and card paths stay in sync?

rishiraj38: Good catch; yes, that was redundant. Updated PermissionShield to pass permissionKey directly to PermissionSessionContext instead of manually deriving everything. Both the tooltip and card paths now go through the same self-resolving logic, so they'll stay in sync automatically.

hiyach28:
Really liked the direction of this refactor. one thing I was curious about while reading it: pages in Meshery still perform their own useHasPermission(...) checks before rendering the error state. so will we continue with it to stay as the current pattern, or do you think it could eventually be abstracted further? that was one of the things I saw earlier, so wanted to understand/ learn from how you think about the long-term architectures.

src/custom/permissions.tsx
```diff
@@ -169,169 +221,208 @@ export const PermissionShield: React.FC<PermissionShieldProps> = ({
...
-            />
-          ))}
-        </Box>
-        <Link
-          href="https://docs.meshery.io/reference/references/permissions/"
-          target="_blank"
-          rel="noopener noreferrer"
-          onClick={(e: React.MouseEvent) => e.stopPropagation()}
+      {/* Bottom row: Category/Subcategory chips + Key Reference link */}
+      {hasKeys && Boolean(categories?.length || subcategories?.length) && (
```
banana-three-join, line 319: In base, the bottom row was rendered unconditionally; only the chip .map() was empty. Now the guard wraps the whole row, so any key lacking category/subcategory loses the docs link entirely.

If this is an unwanted behavior, you can wrap the condition around the child Box component instead. 

```{hasKeys && (
  <Box sx={{ display:'flex', justifyContent:'space-between', ... }}>
    {Boolean(categories?.length || subcategories?.length) && (
      <Box sx={{ display:'flex', gap:0.75, flexWrap:'wrap' }}>…chips…</Box>
    )}
    <Link …>Key Reference</Link>
  </Box>
)}```

banana-three-join (commented):
Other than the small change in behavior pointed out, it looks good to me!

src/custom/permissions.tsx
rishiraj38: Thanks, that was unintended. The "Key Reference" link should always be reachable when keys are displayed. Moved the category/subcategory guard inside the row so it only wraps the chips, and added marginLeft: 'auto' on the link to keep it right-aligned when there are no chips.

banana-three-join (approved):

KhushamBansal (approved):
LGTM!!

PR #1782, opened 2026-08-07
fix: export ActionButton component from sistent

KhushamBansal:
**Notes for Reviewers**

This PR fixes #

## Summary by CodeRabbit

* **New Features**
  * Added the `ActionButton` component to the package’s public exports for use in applications.
  * Added public access to `ActionButtonProps` and `Option` types, improving support for type-safe customization and configuration.

### DIFF
--- a/src/custom/ActionButton/ActionButton.tsx
+++ b/src/custom/ActionButton/ActionButton.tsx
@@ -10,7 +10,7 @@ import {
   Popper
 } from '../../base';
 import { DropDownIcon } from '../../icons';
-interface Option {
+export interface Option {
   icon: React.ReactNode;
   label: string;
   onClick: (event: React.MouseEvent<HTMLLIElement, MouseEvent>, index: number) => void;
@@ -18,7 +18,7 @@ interface Option {
   show?: boolean;
 }
 
-interface ActionButtonProps {
+export interface ActionButtonProps {
   defaultActionClick: () => void;
   defaultActionDisabled?: boolean;
   options: Option[];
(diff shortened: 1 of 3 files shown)


Katotodan (commented):
Adding this export to `src/index.tsx` solves the problem, but I think the code could be cleaner and better scoped if we added the export to `src/custom/index.ts` instead (and still keep the export from `src/custom/index.tsx`).

Currently, it's exported from `src/custom/index.tsx`. If we add the export to `src/custom/index.ts`, I believe it should still work while keeping the change more localized and organized.

I could be mistaken, though.

ishwar170695 (approved):
LGTM. This follows the existing explicit export pattern used in `src/index.tsx` for components affected by the declaration bundling issue.

ishwar170695:
> Adding this export to `src/index.tsx` solves the problem, but I think the code could be cleaner and better scoped if we added the export to `src/custom/index.ts` instead (and still keep the export from `src/custom/index.tsx`).
> 
> Currently, it's exported from `src/custom/index.tsx`. If we add the export to `src/custom/index.ts`, I believe it should still work while keeping the change more localized and organized.
> 
> I could be mistaken, though.

I think the explicit export in `src/index.tsx` is intentional here. The existing comments indicate it's a workaround for `rollup-plugin-dts` dropping declarations behind `export * from './custom'`, so I'd lean toward keeping it consistent with the existing pattern.

banana-three-join (changes requested):
I believe it also needs the typing exported

KhushamBansal:
> I believe it also needs the typing exported

@banana-three-join Thanks for reviewing. I have updated the PR.

banana-three-join (approved):
Ty! LGTM

PR #1807, opened 2026-08-24
[ShareModal] Preserve host visibility error message on failure

### LINKED ISSUE #1789: ShareModal discards the host's visibility-change error text (double .error hop on a string)
## Summary

`ShareModal`'s visibility-change notification always renders as a bare `"Failed to update visibility. "` with no detail, because it dereferences `.error` twice on a value its own prop type declares as a plain string.

## Where

`ShareModal`, in the published `@sistent/sistent@0.22.0` bundle (`dist/index.mjs`):

```js
const B = `Failed to update visibility. ${q?.error?.error || ""}`;
q.error
  ? d({ message: B, event_type: "error" })
  : d({ message: N, event_type: "success" });
```

`q` here is the awaited result of the host-supplied `handleUpdateVisibility` prop.

## Why it is wrong

`ShareModalProps` types that prop as:

```ts
handleUpdateVisibility: (value: string) => Promise<{ error: string }>;
```

So `q.error` is a **string**. The branch test `q.error` is correct, but the message interpolates `q.error.error` - a `.error` lookup on a string - which is always `undefined`, so `|| ""` collapses it to the empty string. Every message a host composes is discarded and the user is told only that something failed, never what.

The adjacent revoke path is fine, since there the value genuinely is an RTK result:

```js
{ error: N?.error?.error }
```

That is likely where the extra hop was copied from.

## Impact

Hosts have no way to surface a reason for a failed visibility change. In Kanvas (layer5labs/meshery-extensions) we normalize RTK's several error shapes into the documented `{ error: string }` and return a specific message for unsupported resource kinds; none of it can reach the user.

This is in the same silent-failure family as the payload casing bug fixed in #1786 - the failure is now correctly *detected* (a 4xx/5xx no longer reports as success), but the explanation is dropped on the floor.

## Suggested fix

```js
const B = `Failed to update visibility. ${q?.error || ""}`;
```

If the double hop is there to tolerate hosts that return an RTK result rather than the documented shape, normalizing once would be clearer than an optional-chained guess:

```js
const detail = typeof q?.error === "string" ? q.error : (q?.error?.error ?? "");
```

## Reproduce

Pass a `handleUpdateVisibility` that resolves to `{ error: "visibility rejected" }` and trigger a visibility change. Expected: the reason appears in the error notification. Actual: `"Failed to update visibility. "`.

## Environment

- `@sistent/sistent` 0.22.0 (verified against the published npm artifact, not a local build)

MAYANKSHARMA01010:
**Notes for Reviewers**

This PR fixes #1789.

**[Signed commits](https://github.com/layer5io/sistent/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**

### Summary of Changes

* **Fix Error Detail Extraction in `ShareModal`**:
  * In `src/custom/ShareModal/ShareModal.tsx`, `notifyVisibilityChange` was performing a double `.error` lookup (`res?.error?.error`) on the result of `handleUpdateVisibility`.
  * Because `handleUpdateVisibility` is typed as `(value: string) => Promise<{ error: string }>`, `res.error` is already a string, causing `.error.error` to evaluate to `undefined` and dropping the error explanation on the floor.
  * Updated `notifyVisibilityChange` to extract `res.error` directly when it is a string, with fallback to nested RTK error structures (`res?.error?.error`, `res?.error?.data?.message`, `res?.error?.message`) for backwards compatibility.
  * Cleaned up fallback message formatting so that no awkward trailing whitespace is emitted when no detail string is provided.
  * Used `!res?.error` safe navigation to prevent runtime `TypeError` when `res` is undefined.

* **Unit Tests**:
  * Added unit test cases in `src/__testing__/ShareModalWireContract.test.tsx` verifying:
    1. String error text is preserved in error notifications (e.g. `'Failed to update visibility. visibility rejected'`).
    2. Nested RTK error text is preserved (e.g. `'Failed to update visibility. permission denied'`).
    3. Success notifications and visibility updates continue to behave as expected.

### Testing Done
- `npm run build` completed successfully with zero type or bundling errors.
- `npm test` passed across all 30 test suites (518 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved visibility-update error handling for different response formats.
  * Prevented incomplete error messages from appearing when details are unavailable.
  * Improved reliability when confirming successful public visibility changes.

* **Tests**
  * Added coverage for successful updates, string errors, nested errors, selected visibility, callbacks, and notifications.


MAYANKSHARMA01010:
@Sbragul26 @KhushamBansal can you review this PR

hiyach28 (commented):
Thank you for your contribution!
Let’s discuss this during tomorrow’s website call at 5:30 PM IST / 6 AM CT.
If you would, please add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1Nys9Ky4XJ1MwoCALJiMtSfQHPkIlGcx5_mITADZVnj4/edit?tab=t.0) :)

PR #1817, opened 2026-08-25
[Tests] Add structural integrity and regression test suite for `src/index.tsx` barrel

rishiraj38:
**Notes for Reviewers**

This PR adds a static structural integrity test suite (`src/__testing__/indexBarrel.test.ts`) for the root package entry point `src/index.tsx`.

### What this test guards against:
1. **Accidental Deletions**: Pinned manifests for both wildcard re-exports (`export * from './...'`) and explicit re-exports (`export { ... } from './...'`). If an export is removed, the test fails and names the exact symbol/specifier missing.
2. **Unpinned Additions**: Two-directional validation ensures any newly added explicit or star re-export is tracked and pinned.
3. **Broken Paths / Typo in Specifiers**: Confirms all `from '...'` specifiers resolve to actual source files.
4. **Source Symbol Renames**: Validates that all explicitly exported symbols exist in their referenced source modules.
5. **Bundle Output Validation**: Verifies that required runtime exports and `MESHERY_EXTENSION_CONTRACT_VERSION` appear in the emitted `dist/` bundle.
6. **Fast Static Analysis**: Operates via static source & bundle analysis (~0.4s execution time), avoiding transitive ESM loading issues with Jest.

## Summary by CodeRabbit

* **Tests**
  * Added automated checks to ensure package exports remain valid, complete, and consistent.
  * Added validation that published build output contains required runtime exports and contract information.
  * CI now detects missing build artifacts and unintended changes to the public API earlier.


ishwar170695 (changes requested):
Requesting changes because the existing CodeRabbit findings are valid and affect the effectiveness of the regression suite, particularly the no-op overlap assertion and incomplete bundle/export validation.

I also found that the source-symbol check only verifies that an identifier occurs somewhere in the file, rather than confirming it is actually exported. Additionally, the regex-based export parsing has syntax blind spots. These checks should be strengthened before relying on the suite as a structural integrity guard.

banana-three-join (commented):
**Found Issues:**

**`PINNED_EXPLICIT_EXPORTS` and `PINNED_STAR_REEXPORTS` are derived from `src/index.tsx`, and both are validated against that same file, in both directions. That closes the loop:**
 
- **Deletions aren't guarded.** Removing an export from the barrel and removing the matching line from the manifest is a two-line diff that turns CI green. The failure message asks the contributor to reconcile the list, not to justify the change.
- **Additions are guarded, but pointlessly.** The `catches unpinned additions` tests mean CI blocks on every new export until the name is pasted into the manifest. Mandatory edit, no signal — the list can never diverge from the file, so it never tells us anything the file doesn't.
The practical outcome is that after the fifth "add the symbol to `PINNED_EXPLICIT_EXPORTS`" failure, people stop reading the message and start pasting — including on the one PR where the removal was accidental.
 
**This doesn't cover the failures we're actually seeing**
 
The breakages that reach consumers are mostly *type* changes, not deleted exports:
 
- a prop becomes required
- a union is narrowed
- a field on an exported interface is renamed
- a generic constraint tightens
In every one of those cases the export name is unchanged, so this suite stays green while consumer builds fail. The file's own header notes that type-only exports are out of scope. That's the majority of our pain.

 **Proposal: typecheck a real consumer against the PR build**
 
```yaml
- uses: actions/checkout@v4
- run: npm ci && npm run build && npm pack
 
- uses: actions/checkout@v4
  with:
    repository: layer5io/meshery
    path: consumer
 
- run: |
    cd consumer/ui
    npm ci
    npm i ../../sistent-*.tgz
    npx tsc --noEmit

PR #1823, opened 2026-08-28
fix(BottomSheet): fallback to background.default and adapt text color

KhushamBansal:
**Notes for Reviewers**

This PR fixes #

Screenshot 2026-08-28 at 8 57 58 PM

## Summary by CodeRabbit

* **New Features**
  * Added customizable header background and text colors for dashboard layouts and mobile bottom sheets.
  * Bottom Sheet headers now use the theme’s surface tint when no custom background is provided.

* **Bug Fixes**
  * Improved header text and close-button contrast across light and dark themes.
  * Corrected fallback behavior for empty or unavailable header backgrounds.
  * Updated default header colors to align with application theme settings.

### DIFF
--- a/src/custom/BottomSheet/BottomSheet.tsx
+++ b/src/custom/BottomSheet/BottomSheet.tsx
@@ -67,8 +67,8 @@ const BottomSheet = ({
               alignItems: 'center',
               padding: '1rem',
               textAlign: 'center',
-              background: headerBackgroundColor || theme.palette.surface.tint,
-              color: headerTextColor || theme.palette.text.primary
+              background: headerBackgroundColor || theme.palette.surface?.tint || theme.palette.background.default,
+              color: headerTextColor || (theme.palette.surface?.tint ? theme.palette.common.white : theme.palette.text.primary)
             })}
           >
             <Typography
@@ -93,7 +93,7 @@ const BottomSheet = ({
               edge="end"
               sx={(theme) => ({
                 '& svg': {
-                  fill: headerTextColor || theme.palette.text.primary
+                  fill: headerTextColor || (theme.palette.surface?.tint ? theme.palette.common.white : theme.palette.text.primary)
                 },
                 transform: 'rotate(-90deg)',
                 '&:hover': {
--- a/src/custom/BottomSheet/BottomSheet.tsx
+++ b/src/custom/BottomSheet/BottomSheet.tsx
@@ -43,7 +43,7 @@ const BottomSheet = ({
 
   const tint = theme.palette.surface?.tint;
   const finalHeaderBackgroundColor =
-    headerBackgroundColor ?? tint ?? theme.palette.background.default;
+    headerBackgroundColor || tint || theme.palette.background.default;
 
   const defaultForeground = headerBackgroundColor
     ? readableTextColor(
--- a/src/custom/BottomSheet/BottomSheet.tsx
+++ b/src/custom/BottomSheet/BottomSheet.tsx
@@ -1,4 +1,5 @@
 import Slide, { SlideProps } from '@mui/material/Slide';
+import { useTheme } from '@mui/material/styles';
 import React, { useId } from 'react';
 import { Box } from '../../base/Box';
 import { Dialog } from '../../base/Dialog';
@@ -38,6 +39,14 @@ const BottomSheet = ({
   headerTextColor
 }: BottomSheetProps) => {
   const titleId = useId();
+  const theme = useTheme();
+
+  const tint = theme.palette.surface?.tint;
+  const finalHeaderBackgroundColor =
+    headerBackgroundColor || tint || theme.palette.background.default;
+  const usingTint = !headerBackgroundColor && Boolean(tint);
+  const finalHeaderTextColor =
+    headerTextColor || (usingTint ? theme.palette.common.white : theme.palette.text.primary);
 
   return (
     <Dialog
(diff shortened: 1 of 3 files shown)


PARTH-TUSSLE (changes requested):
Hey @KhushamBansal , the main thing I noticed is that the text/icon color fallback is checking whether surface.tint exists, rather than the background that actually ends up being rendered.
For example, if someone passes headerBackgroundColor="#fff", we'd still get common.white for the text because surface.tint exists, which gives us white-on-white.
I think we should resolve the final header background first, then derive the default text color from that, while still letting an explicit headerTextColor override it. Also, we could resolve that text color once and reuse it for both the title and close icon instead of duplicating the logic.

KhushamBansal:
> Hey @KhushamBansal , the main thing I noticed is that the text/icon color fallback is checking whether surface.tint exists, rather than the background that actually ends up being rendered. For example, if someone passes headerBackgroundColor="#fff", we'd still get common.white for the text because surface.tint exists, which gives us white-on-white. I think we should resolve the final header background first, then derive the default text color from that, while still letting an explicit headerTextColor override it. Also, we could resolve that text color once and reuse it for both the title and close icon instead of duplicating the logic.


@PARTH-TUSSLE Thanks for reviewing this PR. I have made the change.

src/custom/BottomSheet/BottomSheet.tsx
```diff
@@ -1,4 +1,5 @@
 import Slide, { SlideProps } from '@mui/material/Slide';
+import { useTheme } from '@mui/material/styles';
```
NSTKrishna: use it from ../../theme

NSTKrishna (changes requested):

src/custom/BottomSheet/BottomSheet.tsx
```diff
@@ -38,6 +39,14 @@ const BottomSheet = ({
...
 }: BottomSheetProps) => {
   const titleId = useId();
+  const theme = useTheme();
+
+  const tint = theme.palette.surface?.tint;
+  const finalHeaderBackgroundColor =
+    headerBackgroundColor || tint || theme.palette.background.default;
+  const usingTint = !headerBackgroundColor && Boolean(tint);
+  const finalHeaderTextColor =
+    headerTextColor || (usingTint ? theme.palette.common.white : theme.palette.text.primary);
```
NSTKrishna: it should be theme.palette.text.inverse : theme.palette.text.default

NSTKrishna (changes requested):

NSTKrishna (approved):

PARTH-TUSSLE (changes requested):
Thanks @KhushamBansal! The fallback to `background.default`, sharing the header foreground across the title/close icon, and prop forwarding in `DashboardLayout` look great.

One remaining contrast edge case: if someone passes a dark custom `headerBackgroundColor` (e.g. `#121212`), `usingTint` is `false`,
 which defaults to `text.default` (dark text in light mode), resulting in low contrast.

We can use Sistent's `readableTextColor` helper from `../../theme` so custom backgrounds adapt automatically while keeping explicit
`headerTextColor` as the priority override:

```tsx
    import { useTheme, readableTextColor } from '../../theme';

    // ...
    const tint = theme.palette.surface?.tint;
    const finalHeaderBackgroundColor =
      headerBackgroundColor ?? tint ?? theme.palette.background.default;

    const defaultForeground = headerBackgroundColor
      ? readableTextColor(headerBackgroundColor, theme.palette.text.inverse, theme.palette.text.default)
      : tint ? theme.palette.text.inverse : theme.palette.text.default;

    const finalHeaderTextColor = headerTextColor ?? defaultForeground;
```
Could you also add a few quick unit tests for BottomSheet covering:
  • Default surface.tint fallback
  • Missing tint → background.default fallback
  • Custom light / dark background contrast
  • Explicit headerTextColor override

  Other than that, this looks ready to go!

src/custom/DashboardLayout/DashboardLayout.tsx
```diff
@@ -36,7 +42,9 @@ export const DashboardLayout: React.FC<DashboardLayoutProps> = ({
   sidebarTitle = 'Widget Picker',
   sidebarWidth = { xs: '100%', md: '350px' },
   sidebarTopOffset = '0',
-  sidebarHeight = '100vh'
+  sidebarHeight = '100vh',
```
Katotodan: Would it be better to set `sidebarHeight` to `100dvh` instead of `100vh` to account for dynamic viewport changes?

Katotodan (commented):
Looking good overall, just one suggestion.

ayanasarkar (commented):
verified the color-resolution logic and ran the new test suite locally — all 7 pass, and the dark-mode tint/contrast edge case Parth flagged is correctly handled. Only open item is Katotodan's 100vh → 100dvh suggestion on sidebarHeight, which doesn't look addressed yet — otherwise this looks ready to merge once that's in.

src/custom/DashboardLayout/DashboardLayout.tsx
KhushamBansal: Fixed! Thanks for the review.

Katotodan (approved):

PARTH-TUSSLE (dismissed):
Hey @KhushamBansal ,
1. The empty-string fix wasn't applied symmetrically. The background fallback was hardened to treat "" as absent, but the sibling text-color line still uses ?? So headerTextColor="" would still slip through as an invalid CSS color , silently reproducing the exact bug this PR's last content commit was written to fix , just on the other prop. Small, edge-case, but worth a one-line consistency fix (headerTextColor || defaultForeground) before merge.
2. No regression test for the empty-string case. The test suite (7 tests) is solid , tint fallback in both modes, background.default fallback, light/dark custom-background contrast, explicit override, icon color inheritance , but it doesn't cover the headerBackgroundColor="" scenario, nor the headerTextColor="" gap above. Worth adding one test for each so this doesn't regress silently later.

Other than that LGTM 👍

saurabhraghuvanshii (approved):

PR #1841, opened 2026-09-13
perf: make the published package tree-shakeable

saurabhraghuvanshii:
Consumers could not tree-shake @sistent/sistent at all. Importing one component from 0.22.6 pulls ~14,000 modules and 14.7 MB into a webpack build, 10,775 of those modules being @mui/icons-material. In layer5.io this was the largest single contributor to build memory.

Three causes, all fixed here:

- `module` pointed at the CommonJS build. `main` and `module` were both `./dist/index.js`, so no bundler ever selected `dist/index.mjs`. `module` now points at the ESM build, and an `exports` map routes `import` and `require` to the matching build and declarations.

- There was no `sideEffects` field, so bundlers had to keep every export. The package has no top-level side effects (no CSS imports, no global registration), so it is now marked `sideEffects: false`.

- Six components imported icons from the `@mui/icons-material` barrel, which re-exports ~10,800 modules. They now import by path, and a `no-restricted-imports` rule keeps the barrel out.

Serving the ESM build exposed two imports that only worked through CJS interop, and would have broken strict ESM consumers:

- `import jsyaml from 'js-yaml'` has no default export in js-yaml's ESM build; now `import * as jsyaml`.
- `lodash/debounce` has no extension and lodash has no exports map, which webpack rejects from an `.mjs` file (`fullySpecified`); now `lodash/debounce.js`.

Measured with webpack 5, a consumer that imports only `Box`:

  published 0.22.6    this change*
  modules                       14,132           5,288
  @mui/icons-material           10,775              55
  output                      14.68 MB         2.27 MB

  * with the matching @sistent/mui-datatables deep-import fix (layer5io/mui-datatables), which removes the remaining barrel.

The build succeeds (ESM, CJS and declarations), all 515 tests pass, and lint is clean. The public API is unchanged.

Not changed: `noExternal: [/^@meshery\/schemas/]` (c34bfaac) is what grew the package from 2.3 MB in 0.18.0 to 9.3 MB in 0.18.1, but it was added to fix a schema version mismatch in the Meshery UI, so un-bundling it needs a maintainer decision rather than a drive-by change.

**Notes for Reviewers**

This PR fixes #

## Summary by CodeRabbit

* **Compatibility**
  * Improved package entry-point handling for both ESM and CommonJS consumers.
  * Added explicit package exports and identified the package as side-effect free.

* **Maintenance**
  * Standardized icon imports to dedicated module paths.
  * Updated YAML and utility imports for improved module compatibility.
  * Added a linting rule to prevent barrel imports for Material UI icons.


banana-three-join (approved):

banana-three-join (changes requested):
Build issues in Meshery and Cloud and being investigated...

PR #1846, opened 2026-09-16
fix(dashboard): decouple edit session from sidebar visibility

### LINKED ISSUE #1843: DashboardLayout: sticky sidebar sizes itself from sidebarHeight independently of sidebarTopOffset, so any non-zero offset pushes the panel's bottom off-viewport
## Summary

`DashboardLayout` pins its sticky sidebar at `sidebarTopOffset` but sizes it from `sidebarHeight`, and the two are computed independently. `sidebarHeight` defaults to `100vh`, so **any caller that supplies a top offset and leaves the height alone gets a panel exactly `sidebarTopOffset` taller than the space it can occupy.**

Because the panel is `position: sticky`, that overflow is not recoverable by scrolling: once pinned, its bottom `sidebarTopOffset` worth of content - including the end of its own internal scroll container - sits permanently below the viewport, and its header scrolls up under the page's navbar.

## Where

`DashboardLayout`, the desktop (non-mobile) branch:

```jsx
<Box sx={{ width: sidebarWidth, flexShrink: 0, position: "sticky",
           top: sidebarTopOffset, alignSelf: "flex-start",
           height: sidebarHeight, maxHeight: sidebarHeight }}>
```

with

```ts
sidebarTopOffset = "0",
sidebarHeight = "100vh"
```

The default pair is self-consistent (offset `0`, height `100vh`). Every non-zero offset breaks it, and nothing in the types or at runtime says so.

## Reproduce

Render `DashboardLayout` with `sidebarTopOffset="64px"` and no `sidebarHeight` (the natural call for any app with a fixed navbar), open the sidebar above the `md` breakpoint, and scroll the page down. The sidebar's header leaves the top of the viewport and its last 64px never becomes reachable.

Measured, rather than eyeballed:

```js
const rect = stickyPanel.getBoundingClientRect();
rect.height            // 1000 at a 1000px viewport
window.innerHeight - 64 // 936 - the space actually available
```

## Suggested fix

Derive the default height from the offset the caller already gave:

```ts
sidebarHeight = `calc(100vh - ${sidebarTopOffset})`
```

An explicit `sidebarHeight` still wins, so this is backwards compatible for anyone who already worked around it, and it makes the zero-offset default behave exactly as it does today.

The alternative - documenting that callers must pass both - pushes the navbar height into two places in every consumer, where they can silently drift apart.

## Consumer status

Worked around locally in layer5io/meshery-cloud#5992 by binding the navbar height once and deriving both props from it. That is a correct caller, not a fix: every other consumer that adopts `sidebarTopOffset` will hit the same thing first.

### LINKED ISSUE #1845: DashboardLayout offers no way to reopen the desktop sticky picker: isSidebarOpen is both 'should it exist' and 'is it showing'
## Summary

Below the `md` breakpoint, `DashboardLayout` renders `sidebarContent` inside a `BottomSheet` and pairs it with a floating action button to reopen that sheet. But `WidgetPicker` draws its **own** close control (`aria-label="Close widget picker"`) inside that content, and that control is wired to the consumer's `onClose` - the same callback that clears `isSidebarOpen`.

So on mobile there are two close affordances in the same sheet with different and irreconcilable outcomes:

- **The sheet's own close** clears `DashboardLayout`'s internal open state. `isSidebarOpen` stays true, so the FAB renders and the user can reopen the sheet. Correct.
- **`WidgetPicker`'s X** runs the consumer's `onClose`, which clears `isSidebarOpen`. That unmounts the sheet **and the FAB together**, so the picker cannot be reopened at all until the consumer's own state cycles.

The second one is a dead end the user cannot get out of from the UI.

## Where

`DashboardLayout`, mobile branch (from the built `0.22.6` bundle, names shortened):

```jsx
isSidebarOpen && isMobile && (
  <>
    <BottomSheet open={internalOpen} onClose={() => setInternalOpen(false)} ...>
      {sidebarContent}
    </BottomSheet>
    {!internalOpen && <Fab onClick={() => setInternalOpen(true)} ... />}
  </>
)
```

Both the sheet and the FAB are gated on `isSidebarOpen`, so anything that clears it takes the FAB with it. `WidgetPicker`'s X is rendered inside `sidebarContent`, which is exactly what the sheet is displaying.

## Why the consumer cannot fix this cleanly

A consumer could make its `onClose` breakpoint-aware - clear `isSidebarOpen` on desktop, do nothing on mobile - but that requires restating Sistent's `md` breakpoint in the consumer, so the panel's behaviour and the component's layout decision become two values that must agree and nothing checks. That is the class of bug we were already fixing in layer5io/meshery-cloud#5992, so we declined to introduce another instance of it and left the X consistent with desktop.

## Suggested fix

`WidgetPicker` should not render its own close control when it is being displayed inside the bottom sheet - the sheet already provides one, and the two mean different things. Options, in rough order of preference:

1. `DashboardLayout` tells `WidgetPicker` it is embedded (a context value or an internal prop), and `WidgetPicker` omits its X in that case.
2. `WidgetPicker` takes an explicit prop to suppress the close control, and `DashboardLayout` sets it on the mobile branch.
3. `DashboardLayout` gates the FAB on something other than `isSidebarOpen`, so a consumer clearing that prop still leaves a way back.

Either of the first two keeps the breakpoint knowledge inside the component that already owns it.

## Consumer status

Not blocking. layer5io/meshery-cloud#5992 separated panel visibility from its edit-session lifetime, so the X is no longer destructive there - it closes the panel and the edit session continues with its guards armed. On mobile the picker simply cannot be reopened until edit mode cycles, which is a degraded but non-destructive state, and better than the behaviour it replaced.

Related: layer5io/sistent#1843, layer5io/sistent#1844.

NSTKrishna:
**Notes for Reviewers**

This PR fixes #1845 and fixes #1843

### Summary of Changes
- **Root issue #1845**: `DashboardLayout` previously conflated two distinct concerns into a single `isSidebarOpen` prop: whether the editing session is active vs. whether the sidebar panel is currently visible. On mobile, `WidgetPicker` rendered its own close button that cleared `isSidebarOpen`, unmounting the sidebar and removing the FAB, leaving no way to reopen. On desktop, closing the panel had no reopen affordance.
- **Root issue #1843**: `DashboardLayout` pinned its sticky sidebar at `sidebarTopOffset` while sizing it independently with `sidebarHeight` (default `100vh`). Any non-zero offset pushed the panel's bottom off-viewport. We resolved this by dynamically deriving height as `calc(100dvh - <offset>)` when `sidebarTopOffset` is provided without an explicit `sidebarHeight`.
- **Context introduction**: Added `DashboardLayoutContext` (`isMobile`, `isSheet`, `isSidebarVisible`, `closeSidebar`, `openSidebar`) to propagate layout state down to sidebar content. Exported via `./custom/DashboardLayout` and root re-exported in `src/index.tsx` for declaration bundle preservation.
- **Internal panel visibility state**: Added `isSidebarVisible` to `DashboardLayout` (defaults to `true`). Closing the sidebar panel minimizes it while `isSidebarOpen` remains `true`, keeping the edit session active.
- **Desktop reopen affordance**: When minimized on desktop, rendered a floating action button (FAB) so users can reopen the widget picker without leaving edit mode.
- **Controlled visibility & configuration**: Added optional `sidebarVisible`, `onSidebarVisibilityChange`, and `showReopenFab` props to `DashboardLayout`.
- **WidgetPicker smart close button**: `WidgetPicker` auto-detects its container via context—hiding the redundant close button when embedded in mobile `BottomSheet` (which already has its own close handle) and showing it in desktop sidebar or standalone mode, with an explicit `showCloseButton` override.
- **Context close fallback**: `WidgetPicker` falls back to `context.closeSidebar()` when no `onClose` prop is provided.
- **Test coverage**: Added 30 new unit tests across `src/__testing__/DashboardLayout.test.tsx` and `src/__testing__/WidgetPicker.test.tsx`.

## Summary by CodeRabbit

- **New Features**
  - Added controlled and uncontrolled dashboard sidebar visibility.
  - Added mobile and desktop controls for reopening the sidebar.
  - Added dashboard layout context for sidebar state and open/close actions.
  - Added automatic sidebar sizing based on top offsets.
  - Widget Picker close buttons can now be explicitly shown or hidden and integrate with dashboard sidebar controls.

- **Tests**
  - Added comprehensive coverage for dashboard layout behavior, context integration, sidebar sizing, and Widget Picker interactions.


vedant21-ctr (changes requested):
please consider:

Memoizing the DashboardLayoutContext value to avoid unnecessary consumer re-renders.
Making the reopen FAB aria-label dynamic since sidebarTitle is configurable.
Adding a regression test for repeated renders with isSidebarOpen={false} and changing callback references.
Please confirm/document the intended controlled sidebarVisible behavior across edit-mode transitions.

Otherwise, the #1845/#1843 implementation and test coverage look good.

src/__testing__/DashboardLayout.test.tsx
```diff
@@ -0,0 +1,390 @@
...
+    expect(screen.queryByLabelText('Open Widget Picker')).toBeNull();
+  });
+
+  it('shows FAB when BottomSheet is closed and keeps edit mode active', () => {
+    renderLayout({ isSidebarOpen: true });
+    // BottomSheet has its own close button labelled 'Close'
+    const closeBtn = screen.getByLabelText('Close');
+    fireEvent.click(closeBtn);
+    // FAB appears so the user can reopen (sheet is minimized, but MUI Dialog
+    // keeps DOM content mounted — so we check the FAB, not the close button)
```
Maanvi212006: does MUI Dialog actually keep content mounted when closed? I tested it and the sidebar content does get removed from the DOM once the close transition finishes (Dialog doesn't set `keepMounted`).

src/__testing__/DashboardLayout.test.tsx
```diff
@@ -0,0 +1,390 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import React from 'react';
+import { DashboardLayout } from '../custom/DashboardLayout/DashboardLayout';
+import { DashboardLayoutContext, useDashboardLayoutContext } from '../custom/DashboardLayout/DashboardLayoutContext';
```
Maanvi212006: This import line is over the 100-char printWidth and isn't prettier-formatted. will fail the CI prettier check. please run `format:write`

src/__testing__/DashboardLayout.test.tsx
Maanvi212006: does MUI Dialog actually keep content mounted when closed? I tested it and the sidebar content does get removed from the DOM once the close transition finishes (Dialog doesn't set keepMounted).

NSTKrishna: Yep , the comment was misleading i updated it.

src/custom/DashboardLayout/DashboardLayoutContext.tsx
```diff
@@ -0,0 +1,27 @@
+import { createContext, useContext } from 'react';
```
AnkitRewar11, line 1: Do we really need a new Context here?

`DashboardLayout` is just a generic wrapper for `sidebarContent`, while `WidgetPicker` is a separate component. Adding a Context feels like it adds unnecessary coupling between them and also means adding extra exports to `src/index.tsx`.

Since we usually use Context for more global things like `ThemeProvider` or `PermissionProvider`, I think we can keep this simpler.

src/custom/DashboardLayout/DashboardLayout.tsx
```diff
@@ -42,34 +73,62 @@ export const DashboardLayout: React.FC<DashboardLayoutProps> = ({
...
-  // - set to false when Edit Mode turns OFF
-  // This two-dimension model prevents the sheet from re-opening on every
-  // isSidebarOpen change after the user has intentionally minimized it.
-  const [isSheetVisible, setIsSheetVisible] = useState(isSidebarOpen);
+  // isSidebarVisible represents "panel expanded" (true) vs "panel minimized" (false).
+  // This is independent of isSidebarOpen (the edit-session flag).
+  // In uncontrolled mode, we manage visibility internally.
+  // In controlled mode (sidebarVisible prop provided), the caller drives it.
+  const isControlled = controlledVisible !== undefined;
+  const [internalVisible, setInternalVisible] = useState(defaultSidebarVisible);
```
AnkitRewar11, line 93: Looking at this state logic, do we really need all these extra controlled/uncontrolled visibility props?

`DashboardLayout` already has a simple working pattern with `isSheetVisible` that handles minimize and reopening through the FAB without any sync effects.

We could just rename `isSheetVisible` to `isPanelVisible` and use the same state for desktop as well.

This way:

1. Desktop gets the same minimize/FAB reopen behavior.
2. We can avoid the extra controlled/uncontrolled state logic and `useEffect` syncing.

It would make the component simpler and keep the behavior consistent between mobile and desktop.

src/custom/DashboardLayout/DashboardLayout.tsx
```diff
@@ -111,19 +174,43 @@ export const DashboardLayout: React.FC<DashboardLayoutProps> = ({
...
+            </DashboardLayoutContext.Provider>
+          ) : (
+            showReopenFab && (
+              <Fab
+                color="primary"
+                aria-label="Open Widget Picker"
+                onClick={openSidebar}
+                sx={(fabTheme) => ({
+                  position: 'fixed',
+                  bottom: 24,
```
AnkitRewar11, line 204: On desktop, `bottom: 24, right: 24` puts the FAB in the same bottom-right area where apps usually have floating controls, like Sistent’s `FeedbackButton` or other help widgets. It may not overlap exactly, but it can still feel crowded.

Could we add an optional `fabSx?: SxProps<Theme>` prop to `DashboardLayoutProps`, with the current `bottom: 24, right: 24` as the default? This would let host apps easily move the FAB if that corner is already being used.

src/custom/DashboardLayout/DashboardLayout.tsx
```diff
@@ -42,34 +73,62 @@ export const DashboardLayout: React.FC<DashboardLayoutProps> = ({
...
     prevIsSidebarOpen.current = isSidebarOpen;
-  }, [isSidebarOpen]);
+  }, [isSidebarOpen, setSidebarVisible]);
+
+  const closeSidebar = useCallback(() => setSidebarVisible(false), [setSidebarVisible]);
+  const openSidebar = useCallback(() => setSidebarVisible(true), [setSidebarVisible]);
+
+  // Derive sidebar height: if sidebarTopOffset is a non-zero string or number,
+  // use calc(100dvh - <offset>) so the sidebar never pushes content off viewport.
+  const resolvedSidebarHeight =
```
AnkitRewar11, line 126: Great fix for #1843!

Just one small edge case: if someone passes `sidebarTopOffset="0px"` or `"0rem"`, the current check treats it as non-zero and generates `calc(100dvh - 0px)`. It’s valid CSS, but looks a little unnecessary.

We can use `parseFloat(String(sidebarTopOffset)) !== 0` so all zero-unit values are handled cleanly.

src/__testing__/DashboardLayout.test.tsx
```diff
@@ -0,0 +1,389 @@
...
+
+    // Minimize the sidebar via context
+    fireEvent.click(screen.getByLabelText('minimize'));
+
+    // sidebar is hidden: minimize button gone, reopen FAB visible
+    expect(screen.queryByLabelText('minimize')).toBeNull();
+    expect(screen.queryByLabelText('Open Widget Picker')).not.toBeNull();
+  });
+
+  it('reopens desktop sidebar when FAB is clicked after minimizing', () => {
```
Salmaan-M: The minimize/FAB behavior is covered well here. One missing regression test is the synchronization between `isSidebarOpen` and `sidebarVisible`.

Since the implementation explicitly resets panel visibility when `isSidebarOpen` transitions between edit mode states, could we add a test that exercises those transitions?

For example:

1. Start with `isSidebarOpen={true}` and minimize the panel.
2. Rerender with `isSidebarOpen={false}`.
3. Rerender again with `isSidebarOpen={true}`.
4. Assert that the sidebar is expanded again and the reopen FAB is gone.

This would directly verify the behavior implemented by the `useEffect` and ensure the panel is correctly reset when re-entering edit mode.

src/custom/DashboardLayout/DashboardLayoutContext.tsx
Katotodan: Agree, or as we use Redux in this project.
It could be better to stick on one approach.

